PR #50, opened 2015-01-03
Replace Integer with Long in documentation

gdchamal:
Seem that Integer field no more exists and should be replaced using Long .
Just documentation update to make example work with current master branch status


HonzaKral:
Sorry it was just a case of that field type not being implemented yet, I have to go through the types and make sure we have a class for each.

Issue #55, opened 2015-01-15
Inner exceptions on @property DocType


dourvaris:
Using a property on classes makes debugging a nightmare, for example:
  
```python
from elasticsearch_dsl import DocType, Search
class Foo(DocType):
    @property
    def bars(self):
        return Search().eecute() # with a filter in reality

foo = Foo()
foo.bars
```
This will raise **AttributeError: 'Foo' object has no attribute 'bars'**

The actual exception is **KeyError: "There is no connection with alias 'default'."**
```
  File /elasticsearch-dsl/elasticsearch_dsl/search.py", line 366, in execute
    es = connections.get_connection(self._using)
  File "/elasticsearch-dsl/elasticsearch_dsl/connections.py", line 53, in get_connection
    raise KeyError('There is no connection with alias %r.' % alias)
```

I think the except AttributeError clauses in AttrDict and ObjectBase should be removed, I don't see how they add anything.

dourvaris:
Another problem with properties:
```python
class Foo(DocType):
     _value = 'old'
     @property
     def value(self):
         if self._value is 'old':
             self._value = 'new'
         return self._value
 
Foo().value
>>> 'old'
```

HonzaKral:
They add cosistency - if I am accesing an attribute: `result.title` and the property doesn't exist, I expect an `AttributeError`, not a `KeyError`.

I can see the issues in the properties you wrote here, but I don't see the connection with `AttrDict` and `ObjectBase`, could you please elaborate where these are hurting you? Thanks

dourvaris:
I agree on the consistency aspect, and would definitely expect an `AttributeError` on attribute access, I don't have a good answer there yet, maybe an `AttrDictAttributeError` which will keep the original exception.. I'm not sure if it's possible.

The problem is due to the nature of attrdicts having the tendancy to consider attribute failure as a _missing attribute_ rather than an actual exception. This makes the first access of an attribute on an attrdict the initial error. 

The places where this will bite you in the ass on a complicated model are plenty, a simple case:

```python
from elasticsearch_dsl import DocType
class Product(DocType):
    pass

class Category(DocType):
    @property
    def products(self):
        return Product.all()

category = Category(name='laptop')
category.products
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dan/env/local/lib/python2.7/site-packages/elasticsearch_dsl/document.py", line 166, in __getattr__
[...]

PR #56, opened 2015-01-16
Make AttrDict raise correct exceptions on failing attributes/methods.

dourvaris:


HonzaKral:
This PR doesn't change any behavior - the tests attached pass without the changes in the PR (except on Python 3 where they fail regardless of the patch being applied).

Let's discuss on #55 what it is that is causing the trouble. Thanks!

PR #76, opened 2015-02-18
Add type "completion" in field.FIELDS

blasyrkh123:
According to the http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-suggesters-completion.html we must create field with type 'completion' in order to use Suggesters


HonzaKral:
Thank you!

PR #89, opened 2015-03-04
Added a highlight dictionary to DocType.meta (closes #62)

### LINKED ISSUE #62: When creating a DocType instance from results, save highlights somewhere
I don't think `_meta` is a good option since it won't be accessible in django templates for example.

dcwatson:
Massaging the hit data slightly in `from_es` seemed cleaner than changing the `k.startswith('_')` check to also allow a non-underscored key.


HonzaKral:
Thank you!

We will deal with that part once we deal with the outstanding issue about the metadata.

PR #101, opened 2015-03-14
Update index.rst

amitpandita:
When trying to migrate an aggregation(only) and pass the from_dict , dsl fails and doesn't convert properly until i realized the change in the syntax and added "{'query': {'match_all': {}}}" to the body before calculating the aggregation.

Please correct me if i am missing something here, but i felt that this is something that should be either fixed or be in the documentation.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -179,7 +179,15 @@ existing ``dict``, modifying it using the API and serializing it back to a
 .. code:: python
 
     body = {...} # insert complicated query here
-
+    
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thanks for the report, I think that the correct solution is to instead just allow 'query' to be optional in the dict so I implemented that.

amitpandita:
Thanks for the fix, the idea was to notify you and get it fixed. Glad that it happened this quick :)

PR #105, opened 2015-03-17
Response with no items

mindflayer:
I know we can't cover all the slicing API features, but it could be very useful having a syntax to ask for 0 items.
I use it when I am only interested in aggregations, or I need to know how many items respond to my query.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -148,9 +148,9 @@ class Search(object):
             if n.start and n.start < 0 or n.stop and n.stop < 0:
                 raise ValueError("Search does not support negative slicing.")
             # Elasticsearch won't get all results so we default to size: 10 if
-            # stop not given.
+            # stop not given. [0:0] for an empty Response.
             s._extra['from'] = n.start or 0
-            s._extra['size'] = n.stop - (n.start or 0) if n.stop else 10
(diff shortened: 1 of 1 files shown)


amitpandita:
I think the above feature is already available, provide you use the below syntax

s = Search(using=es, index=settings.ES_INDEX, doc_type="MyDoc").params(search_type="count")

just append "search_type" param to count and you will onle get the aggregations.

mindflayer:
Am I wrong if I say that you are counting the aggregations result? I would like to obtain something similar to the following query:

```{
    "query": {
        "bool": {
            "must": [
                {
                    "term": {
                        "customer_id": "%s"
                    }
                }
            ],
            "must_not": [],
            "should": [
                {
                    "term": {
                        "user": "%s"
                    }
                }
            ],
            "minimum_should_match": 1
        }
    },
    "from": 0,
    "size": 0,
    "sort": [],
    "aggs": {
        "games": {
            "terms": {
                "field": "content_id"
            },
            "aggs": {
                "time_spent": {
                    "sum": {
                        "field": "duration"
                    }
                }
            }
        }
    }
}```

mindflayer:
Yes, you are definitely right. Thank you.

HonzaKral:
Closing this since we have support for this. Thanks @amitpandita for the quick reply.

PR #106, opened 2015-03-20
Add support for passing extra parameters to the query body

suzaku:


HonzaKral:
Hi, thanks for the patch! I am just wondering why you'd use this instead of just using `extra` or, in this case, slicing?

suzaku:
@HonzaKral 

Oh, I didn't notice the `extra` method, thanks for letting me know.

Issue #117, opened 2015-04-07
Filter Aggregation not working


john-parton:
I'm trying to create an aggregation like the one described here: http://www.elastic.co/guide/en/elasticsearch/reference/1.x/search-aggregations-bucket-filter-aggregation.html

```python
agg = A({
    "filter" : { "range" : { "stock" : { "gt" : 0 } } },
    "aggs" : {
        "avg_price" : { "avg" : { "field" : "price" } }
    }
})
agg.to_dict()
{
    "filter": { "range": { "stock": { "gt": 0 } }, 
        "aggs": {
            "avg_price": { "avg": { "field": "price" } }
        }
    }
}
```
The `"aggs"` block is moved inside of the `"filter"` block.

I tried to circumvent `A` like this:
```python
query.aggs.bucket('in_stock_products', {
            "filter" : { "range" : { "stock" : { "gt" : 0 } } },
            "aggs" : {
                "avg_price" : { "avg" : { "field" : "price" } }
            }
        })
query.to_dict()
{
    "query": {
        ...
    }, 
    "aggs": {
        "in_stock_products": {
            "filter": { "range": { "stock": { "gt": 0 } }, 
                "aggs": { "avg_price": { "avg": { "field": "price" } } }
            }
        }
    }
}
```
but it still moved the `"aggs"` block inside of the `"filter"` block. (Probably because the `dict` is converted to the `A` class internally.)

If you try to execute the query, you'll get an error like `QueryParsingException[[catalogue] No filter registered for [aggs]`

I'm using a clone of master.

`0.0.3` doesn't appear to have this problem, but I need support for `Search.suggest`.

I cloned master and did a git revert on this commit: https://github.com/elastic/elasticsearch-dsl-py/commit/856801a8c553adf4f34fe353f6f63cd6cd596126

And it went back to the `0.0.3` (correct) behavior.

john-parton:
I would recommend adding a new test case as well. Something like this:

```python
# test_elasticsearch_dsl.test_aggs
def test_aggregate_filter_invariant():
    d = {
        "filter" : { "range" : { "stock" : { "gt" : 0 } } },
        "aggs" : {
            "avg_price" : { "avg" : { "field" : "price" } }
        }
    }
    a = aggs.A(d)
    assert d == a.to_dict()
```

ourlord:
I guess I am having the same issue with this.
The json query should look like this:

	{
		"query": {
			"query_string": {
[...]

PR #127, opened 2015-04-15
Move `aggs` to root of filter aggregation

mandeburka:


HonzaKral:
Do you have any reasoning for this change? The #117 issue was already solved by https://github.com/elastic/elasticsearch-dsl-py/commit/744a5d945fa99135f0a90bd1adc30b6c88cbb6bc as the test in that commit shows. This change, when applied, doesn't affect the filter agg behavior.

Thanks

mandeburka:
Hi
You are right. Didn't see that change yesterday.

PR #133, opened 2015-04-22
Fix _make_dsl_class with Python 2

benjaminrigaud:
type doesn't like unicode parameter in Python 2


HonzaKral:
Thanks for the quick fix, I need to test with python 2 more.

benjaminrigaud:
Travis could be enough :)

PR #137, opened 2015-04-24
Update my email in AUTHORS

robhudson:


HonzaKral:
Thanks, I took the email from git log, there were two so I apparently chose the wrong one, sorry.

PR #140, opened 2015-04-27
Mark accessing to _meta as expected to fail

pquentin:
https://github.com/elastic/elasticsearch-dsl-py/commit/493c27cec4c1c54ee6b22b26f6dd170a879967c7 does not work as expected: using `hit._meta` no longer works, only `hit.meta` works. Here's a failing test (marked as "expected to fail" with pytest) to show the issue.

### DIFF
--- a/test_elasticsearch_dsl/test_result.py
+++ b/test_elasticsearch_dsl/test_result.py
@@ -1,7 +1,8 @@
-from pytest import raises
+from pytest import raises, mark
 
 from elasticsearch_dsl import result
 
+@mark.xfail
 def test_interactive_helpers(dummy_response):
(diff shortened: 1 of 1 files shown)


HonzaKral:
This was a breaking change with `0.0.4`, the changelog mentions it though maybe not too clearly.

Unfortunately we still need to be breaking a few things since the library is fairly new. All backwards incompatible changes should be marked as such in the changelog though.

pquentin:
Sorry, I had missed the changelog. Thanks! (Separating breaking changes from new features would be helpful. And no worries, we're happy to see the library evolving quickly.)

PR #143, opened 2015-04-27
Adding support for setting specific index_analyzer and search_analyzer properties on string fields.

wkiser:
Tested functionality on my local setup. Added some serialization tests for sanity checking.


HonzaKral:
Thanks for the patch, much appreciated!

PR #145, opened 2015-04-28
If there is no value for a date field, index null instead of raising an error

kaoskitn:
It's a simple little change.  I implemented it by overriding Date, but I think this, or something like it, should be included in ES_DSL.   Thank you!

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -149,6 +149,8 @@ class Date(Field):
     def _to_python(self, data):
         if isinstance(data, date):
             return data
+        elif not data:
+            return None
 
         try:
             # TODO: add format awareness


HonzaKral:
Thanks for the patch. I am not sure about this tbh - it seems to me that you shouldn't have the value in the document at all, if you don't have a value.

What I was thinking about instead if to provide an option like `default=EMPTY|NONE|EXC` to define whether accessing a field that doesn't have a value should return the field's `empty()` value, `None` or raise `AttributeError` (current behavior).

What do you think?

kaoskitn:
I reckon that I would rather see a date field with a value of null than no value for it.  We might want to have something like required/default though.  Required would raise if the value is None, and default would allow it to set a default value.  Good ideas...   In the meantime though, I do think it is valid for a value on a document to be null, which is why I submitted this patch.

HonzaKral:
Well, we could mimic the behavior by allowing the empty value to be returned on field access (very simple change) and I feel like that solution would be cleaner.

In the future I would like to add validation to DocType to allow for things like required and arbitrary rules to determine if a document is valid or not (methods, hooks, validators, ...).

Would that be ok fro your use case or do you actually want a `null` value in elasticsearch?

Thanks for the discussion, it's very useful to have these to make sure we are bulding something that people can use.

kaoskitn:
I'll kick the tests.  I just signed all the docs.   By the way, I'm doing pretty intense work using ES_DSL and I'm probably going to commit other things I find along the way.

kaoskitn:
If it prevents Date._to_python from being called with None as a Date field, then that would be fine with me.

HonzaKral:
So I implemented a first pass in the blank branch: https://github.com/elastic/elasticsearch-dsl-py/compare/blank?expand=1#diff-34d6bbb6c1bc63e482787fa540c41a53R76

It now allows you to specify a field being `blank` which will result in it returning its `empty` value when accessing the value without it being present. This means that you can have the behavior you want without storing `null`s in elasticsearch, Does that make sense?

I don't like the name `blank` for the `Field` parameter and would be super happy if you had a better idea for it. Thanks!

HonzaKral:
This has been fixed by validation on documents.

Thanks again for the patch, please let me know if the validation/blank=True approach doesn't work for you.

PR #152, opened 2015-05-06
Allow passing extra search params through DocType.search

dcwatson:
Turns this:

```python
s = self.document.search()
s._extra = {'track_scores': True}
```

Into this:

```python
s = self.document.search(track_scores=True)
```

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -119,11 +119,12 @@ class DocType(ObjectBase):
         cls._doc_type.init(index, using)
 
     @classmethod
-    def search(cls):
+    def search(cls, **extra):
         return Search(
             using=cls._doc_type.using,
             index=cls._doc_type.index,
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thanks for the patch, I am not sure this is a good idea though. The original line should instead be:

    s = self.document.search().extra(track_strores=True)

This doesn't sound too bad to me. Also adding the `**extra` to search makes little sense to me - most of use cases have no need for `extra` and people would most like expect the kwargs they passed to search to be the same they'd pass into `Search` itself.

Does this make sense?

dcwatson:
Ah, not sure how I missed the .extra() method, thanks. Personally, I see DocType.search() as a shortcut for Search() with using/index/doc_type already filled in, so it seems odd that there's no way to pass through the other __init__ args (extra). But .extra does what I need, so I'm good :)

PR #155, opened 2015-05-08
Only put_settings() when settings change

mdj2:
Currently, if you save a mapping twice, an IllegalOperation exception is raised, *even if there are no changes to the mapping settings*. This PR fixes that problem.

### DIFF
--- a/elasticsearch_dsl/mapping.py
+++ b/elasticsearch_dsl/mapping.py
@@ -1,6 +1,6 @@
 from six import iteritems
 
-from .utils import DslBase
+from .utils import DslBase, stringer
 from .field import InnerObject
 from .connections import connections
 from .exceptions import IllegalOperation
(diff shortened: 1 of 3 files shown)


HonzaKral:
Hi, thanks for the PR!

I agree that the functionality is important but I am not sure if converting to string and comparing that will work. This approach will break for dicts (order of keys is not guaranteed) or when the analysis in elasticsearch contains additional elements.

The better solution I think would be to construct Analyzer objects for the stuff we get from elasticsearch and then do a simple comparison. I think it would even be acceptable to just check what analyzers are not defined and just go by name (at least in version 1).

What do you think?

mdj2:
> I am not sure if converting to string and comparing that will work.

I think you might be misunderstanding what `stringer` does (admittedly, it's a poor name). ES stores primitives as strings. For example, if you set `min_gram=2` in Elasticsearch-dsl land, when you get the analyzer back from ES, the integer 2 is now the string '2'. `stringer` recursively walks a list/tuple/dict and converts non-list/tuple/dict objects to strings.

```python
>>> print(stringer({'a': 1, 2: 3, 'b': [1, 'c', 2.5]}))
{'b': ['1', 'c', '2.5'], 'a': '1', '2': '3'}
```

> This approach will break...when the analysis in elasticsearch contains additional elements

Right. I don't see any problem with ignoring analyzers that already exist in ES and aren't defined by elasticsearch-dsl. That would be a good enhancement.

> The better solution I think would be to construct Analyzer objects for the stuff we get from elasticsearch and then do a simple comparison.

This is just a simple dict comparison (I don't see a clear advantage to constructing Analyzer objects). I can certainly do that if you'd rather. But primitives still need to be converted to strings.

> I think it would even be acceptable to just check what analyzers are not defined and just go by name (at least in version 1)

I don't think silently ignoring changes to configured analyzers would be helpful. I regularly have to adjust my analyzers, and I'd find it rather confusing if save() on a mapping succeeded, but the analysis happening in ES was different.

mdj2:
I'm going to close this. I realize now creating the index and putting the analysis in `Mapping.save` doesn't really make sense. The `_collect_analysis` step will only collect the analysis for that particular mapping, not all the mappings defined using this API.

So if you have two analyzers defined on two different Mappings, and you call `save` on both mappings, only one of the analyzers will be put into ES.

PR #171, opened 2015-06-04
Change document.save/delete to raise ValidationError on no index

adisbladis:
The previous behaviour was incompatible between python versions:

A raise on python3 gives us the following:
```
RuntimeError: No active exception to reraise
```

And on python2:
```
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

This makes the behaviour uniform and the error message more helpful.

### DIFF
--- a/test_elasticsearch_dsl/test_document.py
+++ b/test_elasticsearch_dsl/test_document.py
@@ -265,3 +265,13 @@ def test_meta_fields_can_be_accessed_directly_with_underscore():
     assert md.meta.id == 42
     assert md._id == 42
     assert md.meta.parent is md._parent is p
+
+def test_save_no_index(mock_client):
+    md = MyDoc()
+    with raises(ValidationException):
+        md.save(using='mock')
+
+def test_delete_no_index(mock_client):
+    md = MyDoc()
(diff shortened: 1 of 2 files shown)


test_elasticsearch_dsl/test_document.py
```diff
@@ -1,11 +1,23 @@
...
 
+from elasticsearch_dsl.connections import connections
 from elasticsearch_dsl import document, field, Mapping
 from elasticsearch_dsl.exceptions import ValidationException
 
-from pytest import raises
+from pytest import raises, fixture
+
+@fixture(scope='module')
+def setup_mock_connection(request):
```
HonzaKral: You should be able to just use existing `mock_client` - https://github.com/elastic/elasticsearch-dsl-py/blob/master/test_elasticsearch_dsl/conftest.py#L40

HonzaKral:
Thanks so much for picking this up, this is very useful. I added a comment on a small issue where I think we already have the fixture you need in tests.

adisbladis:
You are absolutely right, I have changed it to use the fixture from conftest.py now.

Thanks for your feedback!

PR #174, opened 2015-06-09
Missing document always raises a NotFoundError exception.

alexandre-figura:
The Elasticsearch client raises a ``NotFoundError`` exception when a document is not found. Hence, it is not relevant to inspect the response body of the server.

Moreover, when a document is not found, the integration tests only verify that the ``NotFoundError`` exception is raised:
https://github.com/elastic/elasticsearch-dsl-py/blob/master/test_elasticsearch_dsl/test_integration/test_document.py#L44

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -136,8 +136,6 @@ class DocType(ObjectBase):
             id=id,
             **kwargs
         )
-        if not doc['found']:
-            return None
         return cls.from_es(doc)
 
     @classmethod


HonzaKral:
Yes, but if you pass in `ignore=404` then the error is not raised and there is an integration test that verifies that this code is actually used:

https://github.com/elastic/elasticsearch-dsl-py/blob/master/test_elasticsearch_dsl/test_integration/test_document.py#L47-L48

It is also documented in the docs in Document Life cycle chapter: https://github.com/elastic/elasticsearch-dsl-py/blob/master/docs/persistence.rst#document-life-cycle

PR #177, opened 2015-06-10
don't fail date field parsing on None

hakanw:


hakanw:
I see now that there's a test for exactly this. But why? How can I allow for a possibly blank date, just like in django with DateTimeField(blank=True) ?

HonzaKral:
There is a solution for that in a branch called `blank` I will try and get it merged and cleaned up asap.

PR #179, opened 2015-06-12
allow a blank date if that has been specified on the date field

hakanw:
I thought this had been fixed with the earlier commit, but apparently the Date error is still there. So here's my proposed commit.


HonzaKral:
This has been changed in e1bf242 where proper validation is starting to take place.

Thanks for your patience.

hakanw:
:+1:

PR #189, opened 2015-06-29
Custom alias and index in DocType.search

smal:
## Proposition to use documents.DocType.search() with custom parameters for Elasticsearch alias and index.

Currently, we can save DocType into any index, as well as get and delete object:
```python

def save(self, using=None, index=None, validate=True, **kwargs):
```

It's strange, why the same behaviour not available for DocType.search().

From my point of view, we have a case, when a large amount data should be loaded and processed in separate index (which is different then in DocType.meta), using custom parameters would allow to keep codebase clean.

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -124,10 +124,10 @@ class DocType(ObjectBase):
         cls._doc_type.init(index, using)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


HonzaKral:
This makes perfect sense, thanks for the patch!

PR #193, opened 2015-07-08
Add missing __all__ lists

### LINKED ISSUE #192: Missing __all__ lists
Some modules like `analysis`, `aggs` or `filters` don't have explicitly defined available classes lists. It's a little hard to figure out which classes are available, when reading the code. Also code completion tools in IDEs are thinking that some classes doesn't exist.

artursmet:
Closes #192


HonzaKral:
Merged as e320876

Thanks!

PR #212, opened 2015-07-22
Closes #211 - Allow additional args to the Search.from_dict function

### LINKED ISSUE #211: Allow additional arguments to the `Search.from_dict` function
Currently, the `Search.from_dict` function only takes a single dictionary and creates a new `Search` object which does not have any knowledge of the client connection, the indicies, etc.  This leads to code similar to the following:

```python
client = Elasticsearch()
s = Search.from_dict({'query': 'blah'})
s._using = client
s._index = ['index1', 'index2']
print(s.execute())
```

The `Search.from_dict` function should allow additional positional and keyword params which it just passes down to the `Search` constructor.  It's a simple fix, I'll have a PR shortly.

djt5019:
This PR addresses #211 by allowing additional query parameters to the `from_dict` function to make the newly created `Search` object more useful.  It may also help save the world, though I'm not 100% on that.

P.S. - I've just signed the CLA.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -162,24 +162,39 @@ class Search(object):
             return s
 
     @classmethod
-    def from_dict(cls, d):
+    def from_dict(cls, d, *args, **kwargs):
         """
-        Construct a `Search` instance from a raw dict containing the search
-        body. Useful when migrating from raw dictionaries.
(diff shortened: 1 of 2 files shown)


HonzaKral:
there is already an API to change all these setting on an existing `Search` object so there is no need to add more params to `from_dict`. Thanks for the patch!

PR #215, opened 2015-07-22
Add required import for example

viksit:
from elasticsearch_dsl import analyzer, tokenizer is missing.


HonzaKral:
This was resolved in 237ee78

`tokenizer` is not used in that snippet.

Thanks for the patch!

PR #217, opened 2015-07-23
fix random score function with s/random/random_score

ngokevin:
In the docs, it's called [random_score](https://www.elastic.co/guide/en/elasticsearch/guide/current/random-scoring.html), not ```random```.

I applied this patch locally, and it gets me past the error ```No function with the name [random] is registered.]``` on ES v1.3.9

### DIFF
--- a/elasticsearch_dsl/function.py
+++ b/elasticsearch_dsl/function.py
@@ -69,18 +69,18 @@ class BoostFactor(ScoreFunction):
             del d[self.name]
         return d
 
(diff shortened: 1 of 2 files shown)


ngokevin:
Hi @HonzaKral, would you mind taking a quick look? Much appreciated!

HonzaKral:
merged as ee5c57a

Thanks for the patch and sorry for the delay!

ngokevin:
thanks!

PR #232, opened 2015-08-04
Add filter scoring function

artursmet:
Allows to use [Function Score filters](https://www.elastic.co/guide/en/elasticsearch/guide/current/function-score-filters.html)


artursmet:
My bad - it's already possible with another syntax - https://github.com/elastic/elasticsearch-dsl-py/blob/master/test_elasticsearch_dsl/test_query.py#L237

PR #234, opened 2015-08-05
Add support for no_stem token filter

### LINKED ISSUE #233: Unable to use 'no_stem' filter in analyzer
I can't use following construction in my analyzer definition:
`analyzer('name', tokenizer='edgeNGram', filter=['standard', 'no_stem'])`

According to documentation, it should be possible:
https://www.elastic.co/guide/en/elasticsearch/guide/current/controlling-stemming.html#preventing-stemming

I think that there is a little bug - missing `no_stem` string in TOKEN_FILTERS definition - https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/analysis.py#L61

artursmet:
Fixes #233


HonzaKral:
`no_stem` is not a built-in token filter in elasticsearch, as discussed in #233

PR #238, opened 2015-08-12
Check if an analysis settings are already in index

jasisz:


mieciu:
I was about to submit an exact fix, but hopefully found this PR !! :+1: 

Cannot wait for this to be merged!

HonzaKral:
Merged as f68aad0

Thanks!

PR #245, opened 2015-08-20
added len method to Response object

ziky90:
I missed possibility to directly call `len()` on response.

This PR implements this functionality.


ziky90:
Is this library alive, is it open to PRs? 
I also don't much understand checks. I've already signed CLA, but it still seems thatI did not and based on the stack trace of travis errors there seems to not pass some tests that are not related to this PR at all.

HonzaKral:
Thanks for the patch!

PR #251, opened 2015-09-07
Added partial_fields support

jesushernandez:
For a given use case, we need to support partial fields, so I decided to add it. I have tried to follow the same style as the rest of the code, but please let me know if you have any comments.

Partial fields are documented here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-fields.html#partial

Cheers

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -321,18 +321,17 @@ class Search(object):
         """
         Control which part of the fields to extract from the `_source` document
 
-        :arg partial: dict specifying which fields to extract from the source
+        :kwargs partial: dict specifying which fields to extract from the source
 
-        ``partial`` should have the format:
-            {
-                'some_id': {
-                        'include': ['field', 'some.nested.field'],
-                        'exclude': ['another_field']
-                },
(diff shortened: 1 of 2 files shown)


elasticsearch_dsl/search.py
```diff
@@ -313,6 +317,30 @@ def fields(self, fields=None):
         s._fields = fields
         return s
 
+    def partial_fields(self, **partial):
+        """
+        Control which part of the fields to extract from the `_source` document
+
+        :arg partial: dict specifying which fields to extract from the source
```
HonzaKral: this is not a single arg, but kwargs instead, could we have examples of what the method call will look, instead of what the dict should be?

HonzaKral:
Thanks, it looks great!

elasticsearch_dsl/search.py
jesushernandez: Sure, updating that

jesushernandez:
I was thinking of including a sample mapping to be used for that query and the response, but felt like the call example is enough?

HonzaKral:
Thank you!

Issue #237, opened 2015-08-11
Unable to use Nested aggregations with facets


artursmet:
There is no support for `Nested` aggregations and filters in faceted search.
I tried to figure out how to implement it, but it looks like quite a big challenge. 
Nested aggregations could contain almost any kind of aggregation, so current implementation of [AGG_TO_FILTER](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/faceted_search.py#L19) and [BUCKET_TO_DATA](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/faceted_search.py#L29) is too simple for it. 

Are there any plans for Nested aggregations/filter support or somebody is already working on it?

HonzaKral:
Good question. I am not sure it is possible to implement this directly - what I would like to do instead is perhaps refactor the class so that it would be easier to add any aggregations manually.

Mostly it would mean creating a method to wrap any aggregation into the appropriate filter (what currently happens in `aggregate`) and also the generating of the responses.

Then you'd just have to add the aggregation by overriding the `aggregate` method and then add an apropriate filter in `filter`.

It is definitely not ideal but I cannot see how we could otherwise afford more advanced aggregations.

What do you think?

Thanks for raising this issue!

solarissmoke:
I've been facing the same issue, and was thinking that one approach could be to move `agg_to_filter()` into the class so that it can be overridden. That way you can decide whether to rely on the default add_to_filter conversions or use your own based on context.

HonzaKral:
So this can now be done with custom code - you need to create a custom `Facet` subclass that will wrap any aggregation you wish to pass in. It might even be possible to provide a NestedMixin for the existing `Facet` classes.  I hope this helps.

pirm-in:
This is actually nice! I've found some problems though, not sure if I am doing something wrong there.

For one thing, I had to create a custom ``Response``-class as well, because the structure to access buckets used in ``FacetedResponse.facet`` won't match the custom structure of course.
And for another thing post-filtering won't work either, because it doesn't know about the nested structure. Any idea how this could be made in a nice way? 

Thanks in advance, I'll try anyways

abhijithmannath:
If you are still stuck with an older version you can use the NestedFacet proxy class in this [https://gist.github.com/abhijithmannath/09e46330689641173537325078647342](gist) to work with nested aggregations

PR #255, opened 2015-09-09
Aggs refactoring

artursmet:
I simplified implementation of Aggregations and removed `agg_to_filter` and `bucket_to_filter` functions.
Now it's easier to add new aggregations or define custom ones (as discussed in #237)


HonzaKral:
Hi, thanks for the patch! I am not ignoring it and I am aware of it, I am just not sure this is the direction we would want to go. But it's just a feeling, no concrete evidence or reasons so I am letting it sit for a while to mull it around.

From the code perspective it is absolutely a major improvement over my code!

HonzaKral:
Thanks again for the PR, I decided to go a different direction in 7e901c1dd9eae6d3c500c88860d7c6833210fc6b

The reason for it is that I didn't want the knowledge to live on aggregations themselves - it would make it hard for users to create theis custom logic and also for us to limit the capabilities.

PR #269, opened 2015-10-02
Add items method to AttrDict

morty:
So that you can do things like:

```
for name, agg in result.aggregations.items():
    ...
```


HonzaKral:
I understand the need for this method, but it is too common for people to have `items` as a key in their documents. That is why we deliberately avoid duplicating all the methods of dict - to not pollute the namespace.

the workaround is:

    for name in result.aggregations:
        agg = result.aggregations[name]
        ...

It is not as elegant of course, but doesn't limit the users' choices.

PR #272, opened 2015-10-08
Check for existing analysis

mieciu:
So that you don't end with "You cannot update analysis configuration on an open index, "

### DIFF
--- a/elasticsearch_dsl/mapping.py
+++ b/elasticsearch_dsl/mapping.py
@@ -69,11 +69,12 @@ class Mapping(object):
         else:
             analysis = self._collect_analysis()
             if analysis:
-                if es.cluster.state(index=index, metric='metadata')['metadata']['indices'][index]['state'] != 'close':
-                    # TODO: check if the analysis config is already there
-                    raise IllegalOperation(
-                        'You cannot update analysis configuration on an open index, you need to close index %s first.' % index)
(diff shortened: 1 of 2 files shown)


elasticsearch_dsl/mapping.py
```diff
@@ -69,11 +69,12 @@ def save(self, index, using='default'):
         else:
             analysis = self._collect_analysis()
             if analysis:
-                if es.cluster.state(index=index, metric='metadata')['metadata']['indices'][index]['state'] != 'close':
-                    # TODO: check if the analysis config is already there
-                    raise IllegalOperation(
-                        'You cannot update analysis configuration on an open index, you need to close index %s first.' % index)
-                es.indices.put_settings(index=index, body={'analysis': analysis})
+                existing_analysis = es.indices.get_settings(index)[index]['settings']['index'].get('analysis', {})
+                if analysis.items() > existing_analysis.items():
```
HonzaKral: I am wondering why the use of the `>` operator, shouldn't we just be looking for a subset?

pdonorio:
This looks like a duplicate of #238, but anyway can I help to make this be merged?

It's very frustrating to define your analyzer at the boot of your application and get the `You cannot update analysis configuration on an open index` if it's not the first time.

fxdgear:
closing

PR #273, opened 2015-10-08
Fix for documentation

morty:
Running the code would result in:

SyntaxError: keyword can't be an expression


HonzaKral:
Thank you!

Issue #279, opened 2015-10-27
iterating over Search (without .execute()) will eat all your memory


starenka:
I suppose this a `ProgrammingError`, nevertheless I think this should be fixed. Basically if you try to iterate over a `Search` object which wasn't executed, it will (I think) stuck in an infinite loop and it's trying to eat all you memory, lunch and then sets your fridge one fire. Take this example:

    from elasticsearch import Elasticsearch
    from elasticsearch_dsl import Search

    c = Elasticsearch()

    c.indices.create(index='my-index', ignore=400)

    c.index(index="my-index", doc_type="test-type", id=42, body={"any": "data", }, refresh=True)
    c.index(index="my-index", doc_type="test-type", id=43, body={"any": "body", }, refresh=True)
    c.index(index="my-index", doc_type="test-type", id=44, body={"any": "time", }, refresh=True)
    c.index(index="my-index", doc_type="test-type", id=45, body={"any": "how", }, refresh=True)
    c.index(index="my-index", doc_type="test-type", id=46, body={"any": "where", }, refresh=True)

    qry = Search(using=c, index='my-index').filter('term', any='time')
    q1 = qry.execute()

    assert len(q1) == 1

    for i, one in enumerate(qry, 1):
        print i, one

running this, you will get:

    (tmp-96bb94e953698a92)starenka /prac/python/elasticsearch-dsl-py master at aae0edf? % /usr/bin/time -v python leak.py
    1 <elasticsearch_dsl.search.Search object at 0x7fcba07f1490>
    2 <elasticsearch_dsl.search.Search object at 0x7fcba07f16d0>
    3 <elasticsearch_dsl.search.Search object at 0x7fcba07f1810>
    4 <elasticsearch_dsl.search.Search object at 0x7fcba07f1950>
    ....
    1360 <elasticsearch_dsl.search.Search object at 0x7fcba05a0990>
    1361 <elasticsearch_dsl.search.Search object at 0x7fcba05a0850>
    1362 <elastics^C
    Traceback (most recent call last):
      File "leak.py", line 24, in <module>
        print i, one
    KeyboardInterrupt
    Command exited with non-zero status 1
            Command being timed: "python leak.py"
            User time (seconds): 0.18
            System time (seconds): 0.02
            Percent of CPU this job got: 16%
            Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.29
            Average shared text size (kbytes): 0
            Average unshared data size (kbytes): 0
[...]

PR #281, opened 2015-10-27
fixes https://github.com/elastic/elasticsearch-dsl-py/issues/279

starenka:
proposing a fix for https://github.com/elastic/elasticsearch-dsl-py/issues/279


marcinn:
I think that `__iter__` should execute search, same as Django`s QuerySet is fetching db rows: https://github.com/django/django/blob/master/django/db/models/query.py#L243

HonzaKral:
We decided on a different approach: 1d41adf

thanks!

Issue #286, opened 2015-11-05
'And' filter outputs incorrect syntax


john-parton:
Here's the current output:

```python
>>> import json
>>> from elasticsearch_dsl.filter import And, Prefix, Range
>>> and_filter = And([Range(postDate={'to': '2010-04-01', 'from': '2010-03-01'}), Prefix(name__second='ba')])
>>> print(json.dumps(and_filter.to_dict(), indent=4))
{
    "and": {
        "filters": [
            {
                "range": {
                    "postDate": {
                        "to": "2010-04-01", 
                        "from": "2010-03-01"
                    }
                }
            }, 
            {
                "prefix": {
                    "name.second": "ba"
                }
            }
        ]
    }
}
```

Output *should* be, as per the documentation at https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-and-filter.html#query-dsl-and-filter
```python
{
    "and" : [
        {
            "range" : {
                "postDate" : {
                    "from" : "2010-03-01",
                    "to" : "2010-04-01"
                }
            }
        },
        {
            "prefix" : { "name.second" : "ba" }
        }
    ]
}
```

I'm pretty sure the `Or` filter has the same issue as well.

HonzaKral:
I don't see the error, could you perhaps provide a test? When I look at the docs page you listed there is a second example that uses the more verbose version with the `filters` key...

john-parton:
I think it's a change that was made in Elasticsearch 2.0.

Here's the link to the more recent documentation: https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-and-query.html

HonzaKral:
The syntax is still valid and works, even with 2.0

PR #287, opened 2015-11-05
Use simplified 'And' filter syntax

john-parton:
Fix for https://github.com/elastic/elasticsearch-dsl-py/issues/286

As far as I can tell, support for the more verbose syntax was removed in Elasticsearch 2.0: https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-or-query.html


HonzaKral:
This is a valid syntax and still works with 2.0.

Though `and` filter has been deprecated with `2.0` anyways.

PR #290, opened 2015-11-09
added convenience fields()/keys() method to the AttrDict

ziky90:
Hi,

When I am working with results I am very much missing something like `keys()` method in the standard Python dict.

To prevent confusion that `AttrDict` is in fact not standard dict I named it for now `fields()`, but I am open to rename it `keys()`, I am open to discussion, I would very appreciate to have this in the library.

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -130,6 +130,9 @@ class AttrDict(object):
     def __iter__(self):
         return iter(self._d_)
 
+    def fields(self):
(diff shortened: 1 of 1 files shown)


HonzaKral:
I am sorry, but this is out of the scope of `AttrDict` - it pollutes the namespace so that the dict wrapped by it couldn't contain key called `keys`. Also - you can just iterate over the `AttrDict` object.

ziky90:
Ok, no problem.

PR #297, opened 2015-11-13
Fix imports in configuration.rst

ColCarroll:
elasticsearch_dsl.connections has no attribute `configure`.  I updated the imports so that the code (and following examples) work as intended.


ColCarroll:
CLA should be all set now.

HonzaKral:
Merged as a8064b0, Thanks!

PR #307, opened 2015-12-01
fix objects iterated incorrectly in Docs

xros:
After doing ```response= s.execute()``` the object response is the object which should be iterated, rather than the object **s**.

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -76,7 +76,7 @@ iterate over the ``Search`` object:
 
 .. code:: python
 
-    for hit in s:
+    for hit in response:
         print(hit.title)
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Hi, Thanks for this PR. Unfortunately it is not correct, the documentation affected is demonstrating new functionality introduced in 1d41adf0 where you actually can iterate over `Search`. The second example shows how you can get the raw **query** to see what is actually being executed, so `s` is indeed correct here.

The changes are in master but not yet released.

xros:
@HonzaKral  Hey, okay I see. Mine was downloaded from pypi directly. Also I agree with the newer usage. So let's see what it has in the next release. :smile:

Issue #247, opened 2015-09-02
Add support for ES 2.0 pipeline aggregations


extesy:
Elasticsearch 2.0-beta now supports parent and sibling aggregations: https://www.elastic.co/blog/out-of-this-world-aggregations and http://amsterdam.luminis.eu/2015/08/31/elasticsearch-2-0-and-pipeline-aggregations/

It looks like I can't use them using elasticsearch-dsl at the moment since `.bucket()` call always creates a nested aggregation while I need a sibling one.

ei-grad:
Just my 50 cents - for now one could just add a sibling aggregation explicitly:
```python
s = Search()
s.aggs.bucket(
    "target", "date_histogram",
    interval="hour", field="@timestamp"
)
s.aggs['target'].bucket(
    "movavg", "moving_avg",
    buckets_path="hau",
    model="holt_winters", window=90 * 24,
    minimize=True, settings={'period': 7 * 24},
    predict=14 * 24,
)
s.aggs['target'].metric(
    "hau", "cardinality", field="distinct_id"
)
```

I doubt how it could be look better in implicit way. May be just add all new aggregations as a bucket aggragations?

HonzaKral:
Fixed by cba1d9e

PR #309, opened 2015-12-12
Add moving_avg aggregation

ei-grad:


ei-grad:
Example usage - https://pub.beakernotebook.com/#/publications/566c36fb-a7c0-4061-b1ca-0b2f38528ea8 , elasticsearch_dsl is amazing :-)

ei-grad:
Probably more specified approach is needed for pipeline aggregations, but I don't think so, see #247

HonzaKral:
Thanks, I implemented all the pipeline aggs in cba1d9e a tiny bit differently, but your patch helped!

PR #320, opened 2016-01-12
added missing comma to frozenset

ppiccolo:


HonzaKral:
Hi, do you have any reason for this change? Throughout the library we typically don't include the last comma in a list and functionally it is not required.

Thanks!

ppiccolo:
I know that the comma is required only for one item set to disambiguate it from an expression, ... however in my case after adding the comma to the frozenset, all works as expected, but at this point maybe that I was inside a IDE caching issue because I'm coming from an library upgrade from 0.0.8 to 0.0.9

PR #322, opened 2016-01-19
Fix Match query serializing function

cranti:
When Match queries have more options than a field name and a query string, the format changes ([documented here](https://www.elastic.co/guide/en/elasticsearch/reference/1.6/query-dsl-match-query.html)). Added a new Query class with a to_dict() method that serializes those queries correctly.

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -1,3 +1,5 @@
+from six import iteritems
+
 from .utils import DslBase, _make_dsl_class
 from .function import SF, ScoreFunction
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thanks for the PR. The original intent was that, if such format was needed, the user would specify it explicitly (which works):

    Match(title={'query': 'some words', 'operator': 'or'})

So far we rely on the user to know the DSL and just translate it to the python syntax and we haven't deviated from that. Mostly because otherwise we would have to document all the options here and the users would have to relearn the DSL in python instead of just simply translating the JSON version into python syntax.

What do you think?

cranti:
That's not clear from the docs, which say that query parameters can all be specified as kwargs, and that the queries will be serialized appropriately. It is nice that there's a workaround right now, but I think it would be better to be able to specify the parameters as kwargs. Seems like sticking everything into a dictionary would remove some of the convenience that the library provides. 
If you don't add this feature, could you update the docs?
Thanks!

HonzaKral:
I understand the frustration, I will update the docs to explicitly state that we are talking only about top-level keys in the query's dictionary.

I am hesitant to change that approach for now - we are still holding on to 1-to-1 mapping from the dsl to the json body. Once we decide to go beyond that we can revisit this issue.

Thanks!

PR #349, opened 2016-02-29
Added _source support

andybarilla:
I needed the ability to do source filtering as documented here: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-request-source-filtering.html

Let me know if it needs tweaking. There's also a merge for master but I figured I'd make sure one of them is correct first.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -386,36 +386,42 @@ class Search(Request):
         s._script_fields.update(kwargs)
         return s
 
-    def source(self, source=None):
+    def source(self, **kwargs):
         """
         Selectively control how the _source field is returned.
 
         :arg source: wildcard string, array of wildcards, or dictionary of includes and excludes
 
         If ``source`` is None, the entire document will be returned for
-        each hit.  If source is a wildcard or array of wildcards, matching
-        fields will be included. If source is a dictionary with keys of 'include' and/or
+        each hit.  If source is a dictionary with keys of 'include' and/or
         'exclude' the fields will be either included or excluded appropriately.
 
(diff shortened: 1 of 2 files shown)


elasticsearch_dsl/search.py
```diff
@@ -382,6 +386,38 @@ def script_fields(self, **kwargs):
...
+            s = Search()
+            s = s.source("obj.*")
+
+            s = Search()
+            s = s.source(["obj1.*", "obj2.*"])
+
+            s = Search()
+            s = s.source({
+                "include": ["obj1.*", "obj2.*"],
+                "exclude": ["*.description"]
```
HonzaKral: this seems cumbersome, how would you feel about using kwargs instead?
```
s = Search()
s = s.source(include=['obj1.*'], exclude=["*.description"])
```
with only the keyword specified being overridden, meaning the example could also be written as:
```
s = Search()
s = s.source(include=['obj1.*']).source(exclude=["*.description"])
```

HonzaKral:
Thanks for the PR, it looks good and it is definitely useful functionality. I left a simple note on the diff.

Also, I would like to start deprecating the `1.x` branch, could you make the PR for master? Thanks!

HonzaKral:
Closing in favor of #352

levylll:
hi, i'd like to use the 'source' in my program, but I get the 'AttributeError: 'Search' object has no attribute 'source'' as my return.

levylll:
@HonzaKral 
hi, i'd like to use the 'source' in my program, but I get the 'AttributeError: 'Search' object has no attribute 'source'' as my return.

PR #351, opened 2016-02-29
Search.delete - delete-by-query

ei-grad:


HonzaKral:
Thanks for the PR. I am just not sure we should expose this functionality since it's possibly huge performance cost - having to unload all the documents (even if just the ids) from elasticsearch and issue a bulk request to delete them. Also, should we try and detect (or require?) the `delete-by-query` plugin (https://www.elastic.co/guide/en/elasticsearch/plugins/2.0/plugins-delete-by-query.html) in elasticsearch itself?

If we decide to implement this feature we'd also need to be wary of parent/child and routing information, including them in the data.

HonzaKral:
Closing this in favor of #597

PR #363, opened 2016-03-12
Update setup.py

lowks:
Adding `description` to `setup.py`


HonzaKral:
There is a description already in setup.py a few lines higher. This PR makes the setup.py invalid.

PR #364, opened 2016-03-14
Typo/rst fixes

3lnc:


docs/search_dsl.rst
```diff
@@ -391,7 +391,7 @@ To specify a suggest request on your ``Search`` object use the ``suggest`` metho
 
 .. code:: python
 
-    s = s.suggest('my_suggestion', 'pyhton', term={'field': 'title'})
+    s = s.suggest('my_suggestion', 'python', term={'field': 'title'})
```
HonzaKral: this misspelling has actually been on purpose to show the typical use case for suggestions - the "did you mean?" spell checker :)

3lnc: Whoops. Saw no mention about that )
Rolled back to original.

HonzaKral:
Thanks for the patch and for the change, it is very much appreciated!

PR #366, opened 2016-03-15
Fixed typo in __all__

3lnc:
Idk if we need such test at all.

### DIFF
--- a/test_elasticsearch_dsl/test_aggs.py
+++ b/test_elasticsearch_dsl/test_aggs.py
@@ -166,3 +166,8 @@ def test_filters_correctly_identifies_the_hash():
         }
     } == a.to_dict()
     assert a.filters.group_a == query.Q('term', group='a')
+
+def test_import__all__():
+    for name in aggs.__all__:
+        exec('from elasticsearch_dsl.aggs import {0}'.format(name))
+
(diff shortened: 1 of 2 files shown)


test_elasticsearch_dsl/test_aggs.py
```diff
@@ -166,3 +166,7 @@ def test_filters_correctly_identifies_the_hash():
         }
     } == a.to_dict()
     assert a.filters.group_a == query.Q('term', group='a')
+
+def test_import__all__():
+    for name in aggs.__all__:
+        exec('from elasticsearch_dsl.aggs import {0}'.format(name))
```
HonzaKral: I think `assert hasattr(aggs, name)` would be better here, `exec` is scary even in tests :)

HonzaKral:
I think tests are great but I left a comment on the implmentation.

Thanks for another PR!

test_elasticsearch_dsl/test_aggs.py
3lnc: Indeed its scary )
Damn, haven't figured out such a simple way to test. TY, fixed.

HonzaKral:
Thank you!

PR #369, opened 2016-03-16
Support multi get

brainix:
I love that you can call `DocType.get()`.  So I tried to follow the design there to implement `DocType.mget()`.

@HonzaKral - tagging you so that you get a notification.  Please let me know if I need to change anything?

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -1,5 +1,6 @@
 import re
 
+from elasticsearch.exceptions import NotFoundError, RequestError
 from six import iteritems, add_metaclass
 
 from .field import Field
@@ -148,6 +149,52 @@ class DocType(ObjectBase):
         if not doc['found']:
             return None
         return cls.from_es(doc)
+
+    @classmethod
+    def mget(cls, docs, using=None, index=None, raise_on_error=True,
+             missing='none', **kwargs):
+        if missing not in ('raise', 'skip', 'none'):
+            raise ValueError("'missing' must be 'raise', 'skip', or 'none'.")
+        es = connections.get_connection(using or cls._doc_type.using)
+        body = {'docs': [doc if isinstance(doc, dict) else {'_id': doc}
+                         for doc in docs]}
+        results = es.mget(
+            body,
+            index=index or cls._doc_type.index,
+            doc_type=cls._doc_type.name,
+            **kwargs
+        )
+
+        objs, error_docs, missing_docs = [], [], []
+        for doc in results['docs']:
+            if doc.get('error'):
+                error_docs.append(doc)
+                if missing == 'none':
+                    objs.append(None)
+            elif doc.get('found'):
+                if (raise_on_error and error_docs or
+                    missing == 'raise' and missing_docs):
+                    # We're going to raise an exception anyway, so avoid an
+                    # expensive call to cls.from_es().
+                    pass
+                else:
+                    objs.append(cls.from_es(doc))
+            else:
+                # The doc didn't cause an error, but the doc also wasn't found.
+                missing_docs.append(doc)
+                if missing == 'none':
+                    objs.append(None)
+
+        if raise_on_error and error_docs:
+            error_ids = [doc['_id'] for doc in error_docs]
+            message = 'Required routing/parent not provided for documents %s.'
+            message %= ', '.join(error_ids)
+            raise RequestError(400, message, error_docs)
+        if missing == 'raise' and missing_docs:
+            missing_ids = [doc['_id'] for doc in missing_docs]
+            message = 'Documents %s not found.' % ', '.join(missing_ids)
+            raise NotFoundError(404, message, missing_docs)
+        return objs
 
     @classmethod
     def from_es(cls, hit):
--- a/test_elasticsearch_dsl/test_integration/test_document.py
+++ b/test_elasticsearch_dsl/test_integration/test_document.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from pytz import timezone
 
-from elasticsearch import ConflictError, NotFoundError
+from elasticsearch import ConflictError, NotFoundError, RequestError
 
 from elasticsearch_dsl import DocType, Date, String, construct_field, Mapping
 from elasticsearch_dsl.utils import AttrList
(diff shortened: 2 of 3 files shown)


brainix:
Ah, apologies.  I just signed the contributor agreement.

HonzaKral:
Thanks, this is definitely a great addition to the library!

There are just some issues we need to work out, I am happy to do it, or let me know if you'd be interested in working on it.

- [ ] tests
- [ ] handling missing docs (implement `RAISE`, `SKIP` and `NONE` to control if an exception should be raised, not found doc omitted or returned as None)
- [ ] documentation

We should also use the short version of the multi get API where `index` and `doc_type` are in the url and body only contains `ids` (see last example above https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html#mget-type)

brainix:
Thank you, @HonzaKral, for the insightful feedback, and also for the rest of your hard work.  As you can see in this PR, I believe that I've already done the last part (using the short version of the multi get API).

I believe that I can also handle the `raise`, `skip`, or `none` portion.  Regarding the implementation, I'm thinking about adding a parameter, `missing`, which accepts the values `'raise'`, `'skip'`, or `'none'`, branches accordingly, and throws a `ValueError` is any other value is passed.  Does this approach seem reasonable?

I can also take a crack at updating the docs, but I could definitely use your insight/help with testing.  Awaiting your further input, and thanks again!

brainix:
@HonzaKral: I've updated this PR to handle the `missing` parameter, as well as to include certain tests.  AFAIK, the only thing remaining is documentation.  I'm not the strongest documentation writer, so would you be willing to help with this piece?  And please let me know if the rest looks ok?  If so, I can squash my commits.  Thanks!

HonzaKral:
Thank you for the patch, don't worry about the docs, I am more than happy to write the docs.

The rest looks good with just one tiny detail - the exception. It feels like it should extend the `NotFoundError` from `elasticsearch-py`, or we can just raise that exception directly.

Note that if we raise it directly we need to follow the API by pasing in `404, "Document %s not found." % doc['_id'], doc`

Also the question if `raise` behavior is active - do we raise just for the first document or do we go through the whole list to find all the documents that are missing and raise one "compound" exception.

What do you think?

brainix:
Thank you for the insightful feedback as always, @HonzaKral.  I'm now raising the correct exception, and I chose your "compound" exception idea.  Here's my thinking:

1. If a user passes in a list of IDs into `mget()`, it may be a bit misleading to raise on only the first missing document.
2. While it's more Pythonic to do things lazily/dynamically and to not keep too much in memory, in the case of `mget()`, we have to construct a list anyway.  So I'm not as concerned with additionally constructing a list of missing IDs in order to raise the compound exception.

Is my thinking sound?  Am I missing anything?

elasticsearch_dsl/document.py
```diff
@@ -150,6 +151,32 @@ def get(cls, id, using=None, index=None, **kwargs):
...
+        docs = es.mget(
+            {'ids': ids},
+            index=index,
+            doc_type=cls._doc_type.name,
+            **kwargs
+        )
+        objs, missing_ids = [], []
+        for doc in docs['docs']:
+            if doc['found']:
+                objs.append(cls.from_es(doc))
```
HonzaKral: can we skip calling `from_es` if we know we will be raising `NotFoundError` anyway? it can be quite costly. Doing:
```
if doc['found'] and not missing_ids:
    objs.append(...)
elif not doc['found']:
    ....
```
Should do it

elasticsearch_dsl/document.py
```diff
@@ -150,6 +151,32 @@ def get(cls, id, using=None, index=None, **kwargs):
...
+            else:
+                if missing == 'none':
+                    objs.append(None)
+                elif missing == 'raise':
+                    missing_ids.append(doc['_id'])
+        if missing == 'raise' and missing_ids:
+            message = 'Documents {} not found.'.format(', '.join(missing_ids))
+            raise NotFoundError(404, message)
```
HonzaKral: I would love to see the actual missing docs to be added as the third parameter - a list of the missing records, not just the IDs - do `missing_ids.append(doc)` and raise `NotFoundError(404, message, missing_ids)`

test_elasticsearch_dsl/test_integration/test_document.py
```diff
@@ -60,6 +60,39 @@ def test_get_with_tz_date(data_client):
...
+
+def test_mget_missing_docs(data_client):
+    commit_ids = [
+        '0',
+        '3ca6e1e73a071a705b4babd2f581c91a2a3e5037',
+        'eb3e543323f189fd7b698e66295427204fff5755',
+    ]
+
+    with raises(ValueError):
+        Commit.mget(commit_ids, missing='raj')
```
HonzaKral: can we split all of these into individual tests? that way, when one fails, we know where the error was instead of just knowing that something is wrong. I think this should be 4 tests instead of 2 (test with missing='none' is already implemented in `test_mget`, no need to duplicate it)

HonzaKral:
I really like your solution, it makes perfect sense to me. I left a few nitpicky comments on the diff, if you could look at them and then squash this PR we are good to go.

Thanks for the patch and for the quick responses!

brainix:
Hi, @HonzaKral.  Your comments aren't nitpicky - 1,000 implementation details have to come together under a coherent design in order to make a great library!

I believe that I've addressed all of your feedback, and I've squashed my commits.  Please take a look and let me know if everything is ok.  Thanks!

HonzaKral:
I seem to have forgotten an important part - sometimes passing in `ids` is not enough - when you for example have parent/child or required routing. I think we need to be more flexible in this.

My apologies for not catching this sooner so we could have dealt with it all at once. The travis suite was broken (not starting elasticsearch properly so all integration tests were skipped, now fixed in ce8c80155eeabaa07d62cd7bed18be8c3fc3c02b, rebase to see travis-ci fail) so I assumed all the tests would pass.

I think the best option is to allow people to pass in a list of ids or dicts, so eesentially what we'll do is:
```
body = {"docs": [doc if isinstance(doc, dict) else {"_id": doc} for doc in ids]}
```

Or do you think having two args (`ids` and `docs`) and forcing the people to use one of them explicitly is better? I am torn on this.

Also to catch this we should have a test that produces a failure (required routing/parent not provided - since our test data have parent/child this is easy) because then we will see that we need to check `doc['error']` as well and also probably introduce another parameter - `raise_on_error` (defaulting to `True`) that, if it encounters an error (`if doc['error']`), will raise an exception appropriate to the error. If `raise_on_error` is `False` we will treat errored documents as missing.

brainix:
Great catch, @HonzaKral, and thanks for making the integration tests run again!  I believe that I've addressed your feedback with my latest commit.  I've also taken the liberty to refactor the code to make the logic easier to understand, especially since we're branching on both `raise_on_error` and `missing`.  Please take a look, and let me know if it's ok?

HonzaKral:
Thanks! One last change - when an error happens we shouldn't raise `NotFoundError` with 404, instead probably a `RequestError` is better with code 400. Then if you'd squash again I will merge it.

brainix:
@HonzaKral: Done!  Please take a look and let me know if I've missed something?

brainix:
Hi, @HonzaKral!  Just pinging you - does this PR look ok now?

HonzaKral:
It looks great, thanks again for all the work, will merge shortly as soon as I write the docs.

JakeAustwick:
This looks awesome, thanks @brainix. Can't wait for it to get merged :).

HonzaKral:
Merged as 2564363, sorry for the delay!

I added some docs in 7d3a1b2 and also applied some cosmetic changes to make the flow a little flatter.

Thanks again for the patch!

PR #376, opened 2016-03-25
Add retry_on_conflict META field to doc update

bofia:


HonzaKral:
Hi, thanks for the PR, might I ask what is the motivation behind adding the key to the set?

If you wish to specify `retry_on_conflict` when updating I think it would be cleaner to add it as a keyword argument to the `update` method since it is not really a meta field on the document in the same way as ttl and timestamp are.

what do you think?

PR #423, opened 2016-06-28
Fix suggest() example in persistence doc

dumpster-fire:
s.suggest() returns an elasticsearch_dsl.search.Search object, which is what you want to run execute_suggest() on. 

The previous implementation of this example doesn't store this object, so the example doesn't work (title_suggestions will not exist), and the program will crash.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -318,7 +318,7 @@ If you want to run suggestions, just use the ``suggest`` method on the
 .. code:: python
 
     s = Post.search()
-    s.suggest('title_suggestions', 'pyth', completion={'field': 'title_suggest'})
+    s = s.suggest('title_suggestions', 'pyth', completion={'field': 'title_suggest'})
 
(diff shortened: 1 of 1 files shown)


prashker:
:+1:

HonzaKral:
Thanks for the PR, good catch!

Could I please ask you to sign our CLA (0) so that it can be merged?

dumpster-fire:
@HonzaKral I would if I could! 

When I click "sign agreement" the page refreshes and nothing happens. Spooky.

HonzaKral:
Oops, our bad, sorry about that! Something is broken on the CLA page, thanks for the report!

We are looking into it, I will comment again once we fix it.

prashker:
@dumpster-fire agreement seems to work again

dumpster-fire:
Thanks @prashker 

@HonzaKral I signed it!

HonzaKral:
Thanks!

PR #426, opened 2016-07-07
DocType.update: added detect_noop and doc_as_upsert support

AndreCimander:
Feel free to use, we redirect .save() to update() with upsert to avoid a careless calls to .save() overwriting precious data =)


elasticsearch_dsl/document.py
```diff
@@ -262,7 +265,15 @@ def to_dict(self, include_meta=False):
...
 
-    def update(self, using=None, index=None, **fields):
+    def update(self, using=None, index=None, detect_noop=True, doc_as_upsert=False, **fields):
+        """ Update this document in Elasticsearch.
+
+        :param using:
+        :param index:
+        :param detect_noop: check if values have changed to avoid unnecessary write operation. Default: True.
+        :param doc_as_upsert: index document if it doesn't exist. Default: False.
+        :param fields: kwargs of additional fields
```
HonzaKral: Thanks for the patch, in the rest of the code we use `:arg NAME:` syntax instead of `:param NAME:`, could you please change it to be consistent? Thanks!

AndreCimander:
Oh frack, it's been a while since the 15th of July... sorry for the extreme delay, I had this PR on my to-do list but high prio things just kept piling up, we had some serious crunch time over the last few months *sigh*

As a bonus: I've added the ability to use update() without any fields in order to send the whole document and added field serialization via to_dict().

I'll check the failed travis build and update the PR accordingly.

karmi:
Hi @AndreCimander, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/426.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

AndreCimander:
@karmi hm, didn't I use my googlemail.com address to sign the elasticsearch CLA? At least my inbox says so.

I added my work email as well, just in case (as I just started to move additional changes from our codebase  into my fork).

elasticsearch_dsl/document.py
```diff
@@ -277,25 +280,50 @@ def to_dict(self, include_meta=False):
...
         if not fields:
             raise IllegalOperation('You cannot call update() without updating individual fields. '
                                    'If you wish to update the entire object use save().')
+
         es = self._get_connection(using)
 
-        # update the data locally
-        merge(self._d_, fields)
+        # update given fields locally
+        if len(fields):
```
HonzaKral: no need for this, `IllegalOperation` would have been raised if no `fields` given

elasticsearch_dsl/document.py
```diff
@@ -277,25 +280,50 @@ def to_dict(self, include_meta=False):
...
+            merge(self._d_, fields)
+
+        # prepare data for ES
+        doc = self.to_dict()
+
+        # if fields were given: partial update
+        if fields:
+            doc = dict(
+                (k, doc.get(k))
```
HonzaKral: This would break for dotted fields like `name.first`

Edit: this is not supported anyway, please ignore ;)

HonzaKral: What is the reason for changing the logic and instead of just sending the fields trying to extract the fields from the `doc`? Did you find a case where the original approach didn't work?

AndreCimander: yeah, we had a case where the field serialization wasn't applied after updating the DocType's values by calling merge(self._d_, fields).

If I recall it correctly it was a datetime/epoch_second issue, but that was quite a while ago, so I might err on this.

elasticsearch_dsl/document.py
AndreCimander: hm, don't we want to be able to assign values to the DocType instance and then simply call .update()  to issue a full update?

HonzaKral: No, that's why we have `save()`, `update` is purely for partial explicit updates.

elasticsearch_dsl/document.py
```diff
@@ -277,25 +280,50 @@ def to_dict(self, include_meta=False):
...
         # extract parent, routing etc from meta
         doc_meta = dict(
             (k, self.meta[k])
             for k in DOC_META_FIELDS
             if k in self.meta
         )
+        body = {'doc': doc}
+        if doc_as_upsert:
+            body['doc_as_upsert'] = True
+        body['detect_noop'] = True if detect_noop else False
```
HonzaKral: `detect_noop` should already be a `bool`, do not serialize it again, same for `doc_as_upsert` - just assign those values to the `body`

elasticsearch_dsl/document.py
HonzaKral: I see, this makes sense, just please change the logic to be simpler since we know that we will always have `fields` specified.

fxdgear:
@AndreCimander would you be willing to continue this PR? There are some merge conflicts that need to be resolved still.

PR #432, opened 2016-07-20
Allow custom analyzers to be set from DocType Meta

shezadkhan137:
This allows you to set custom analyzers for a Index through the DocType Meta. Example

```python
html_strip = analyzer('html_strip',
    tokenizer="standard",
    filter=["standard", "lowercase", "stop", "snowball"],
    char_filter=["html_strip"]
)


class Post(DocType):
    title = String()

    class Meta:
        index = 'blog'
        analyzers = (html_strip,)
```

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -378,7 +378,7 @@ in a migration:
 
 .. code:: python
 
-    from elasticsearch_dsl import Index, DocType, String
+    from elasticsearch_dsl import Index, DocType, String, analyzer
 
     blogs = Index('blogs')
 
@@ -400,6 +400,16 @@ in a migration:
     @blogs.doc_type
     class Post(DocType):
         title = String()
+
+    # You can attach custom analyzers to the index
+
+    html_strip = analyzer('html_strip',
+        tokenizer="standard",
+        filter=["standard", "lowercase", "stop", "snowball"],
+        char_filter=["html_strip"]
+    )
+
+    blog.analyzer(html_strip)
 
     # delete the index, ignore if it doesn't exist
     blogs.delete(ignore=404)
--- a/elasticsearch_dsl/index.py
+++ b/elasticsearch_dsl/index.py
@@ -9,6 +9,7 @@ class Index(object):
         self._using = using
         self._settings = {}
         self._aliases = {}
+        self._analysis = {}
 
     def clone(self, name, using=None):
         i = Index(name, using=using or self._using)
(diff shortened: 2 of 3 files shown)


HonzaKral:
Thank you for the patch. I am afraid `DocType` `Meta` isn't the best place for this - it would make much more sense as a method on the `Index` object like `aliases` or `settings`.

Either way I am curious what is the use case for defining an analyzer without using it?

Thanks!

karmi:
Hi @shezadkhan137, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/432.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

shezadkhan137:
@HonzaKral Thanks for the quick reply! Our use case is we have some custom analyzers defined in  our code, however some of the analyzers are used only as "search_analyzers" in a query. So they are not created on DocType.init(). Eg:

```python
common = Common(description={"query": search_query, "analyzer": "description_analyzer"})
match_phrase = MatchPhrase(description={"query": search_query, "boost": 2, "analyzer":"common_search_description_analyzer"})
b = Bool(must=common, should=match_phrase)
...
```

We would like the analyzers to be created when we call DocType.init(), and not have to add the analyzers manually to all our environments.

HonzaKral:
of course, makes sense.

I still think it would be better on the `Index` object where it belongs since it's an index-level setting in elasticsearch too.

something like:
```
i = Index("my-index")
i.settings(number_of_shards=1, number_of_replicas=0)
i.analyzer(html_strip)
i.doc_type(MyDocType)
i.create()
```
What do you think?

shezadkhan137:
Ok,  @HonzaKral I've changed it to work like you suggest on the Index class, as I think you are right, it is more flexible in usage.

shezadkhan137:
@HonzaKral how does this look to you?

HonzaKral:
Thank you @shezadkhan137, this looks great. We also need some tests and some docs to merge this, but I can write those if you prefer.

Thank you also for the ping and your patience!

shezadkhan137:
@HonzaKral no worries, I will do a few tests and try and update the docs.

shezadkhan137:
@HonzaKral I've added a couple of tests and docs. How does this look to you?

shezadkhan137:
@HonzaKral how does this look to you? Is this good to merge?

HonzaKral:
Merged as 05a23cb

Thank you for the work and the patience!

PR #451, opened 2016-08-30
Update persistence.rst

chaopli:
I ran into an exception when I used the constructor in the Nested object. This change could work.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -26,8 +26,8 @@ The mapping definition follows a similar pattern to the query dsl:
 
     # you can also create a field manually
     comment = Nested()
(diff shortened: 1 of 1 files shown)


chaopli:
Signed

HonzaKral:
Thanks for the patch @chaopli; I am, however, unable to reproduce any problem with the original code as
```
from elasticsearch_dsl import Nested, String
n = Nested()
n.field('description', String())
```
works just fine for me. Could you please describe the issue you are trying to solve?

Thanks!

chaopli:
I am not using the code I came across the issue. I will open it again once I come across the issue again.

PR #467, opened 2016-10-06
Correction for displaying python code

peterbe:
Or else it looks like this:
screen shot 2016-10-06 at 8 44 14 am


HonzaKral:
Thanks for the fix!

PR #469, opened 2016-10-06
Type check more safely

brainix:
`isinstance(var, dict)` is True only when `var` is a `dict`; not when `var` is a `collections.OrderedDict` or a `types.MappingProxyType`, for example.

### DIFF
--- a/elasticsearch_dsl/aggs.py
+++ b/elasticsearch_dsl/aggs.py
@@ -1,3 +1,5 @@
+import collections
+
 from .utils import DslBase, _make_dsl_class
 
 __all__ = [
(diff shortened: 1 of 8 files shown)


brainix:
Hi, @HonzaKral!

Tagging you so that you get a notification.  In Python 3, I like to use `types.MappingProxyType` instead of `dict`s for default values for keyword arguments.  The advantage is that `types.MappingProxyType` is like an immutable dict, and it's good practice to use only immutable default values for keyword arguments.

If the code here is convoluted or doesn't make sense for other people's use cases, then I understand!  But I thought I'd submit it anyway in case you feel the same way as I do about type checking in Python.

Thanks,
Raj

HonzaKral:
Thank you for the patch, it has been merged as 420c60d7a59ef843c72c92b675c8cf4b5d0f928d

PR #473, opened 2016-10-12
Fix RangeFacet value filtering

noirbizarre:
According to [official documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html) (and to usage), Range Query expect `gte` and `lt` (to be consistent with the aggregation) instead of `from` and `to` (which are only used in aggregation).


HonzaKral:
Thanks for the patch~

PR #474, opened 2016-10-12
Ensure empty fields are cloned

noirbizarre:
Right now, when `Search._fields` is set to be empty (ie. `s.fields([])`), it's lost on clone and it's not posssible to only require metadata.
This pull-request fix this case.


HonzaKral:
Thank you for the patch, it as been merged into the `2.x` branch

PR #479, opened 2016-10-20
Remove errant colon that broke serialization code block formatting

arusahni:
The current docs had an incorrectly-formatted code block due to an extra colon hanging around.


arusahni:
I just signed the CLA. Is it possible to re-run the check?

PR #487, opened 2016-10-27
Require elasticsearch-py 5.0.0

brainix:

PR #495, opened 2016-11-04
remove restriction elasticsearch <3.0.0

vincentfretin:
remove restriction elasticsearch <3.0.0 to be able to use it with elasticsearch 5.0.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ f.close()
 install_requires = [
     'six',
     'python-dateutil',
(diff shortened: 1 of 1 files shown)


HonzaKral:
Master is not compatible with `5.0`, only the 5.x branch is so I fixed the dependencies there - 4622a2661c1406ffc4016c53c907bd8caef0e68f

Thank you for the report!

vincentfretin:
There was actually another PR #487 with exactly what you committed. You can close it too.

PR #499, opened 2016-11-07
Move default value of 'expand__to_dot' to a variable

### LINKED ISSUE #498: Move _expand__to_dot's default value to a variable
In DslBase the _expand__to_dot_ variable is by default set to True.
It's troublesome in mappings like mine, where I'm using double underscores regurarly in field names and it would be much easier to just set the default value once and forget about it.

Plus, after creating a query that had it set to False it's hard to use your operands on queries as they're just creating a new one with no option to pass that param. 
What's more, even after joining queries 'manually' I can't run them since assignment of a query to Search's query is done by ProxyDescriptor that - in effect - makes it impossible to use your library with field names like mine.

If I'm wrong and I am actually able to run such a query from the Search class - how can I do it? And even if I can; I'd still very much prefer to set the variable once and forget about it instead of passing it each time.

TheName:
Closes #498

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -6,6 +6,7 @@ from six.moves import map
 from .exceptions import UnknownDslObject, ValidationException
 
 SKIP_VALUES = ('', None)
+EXPAND__TO_DOT=True
 
 def _wrap(val, obj_wrapper=None):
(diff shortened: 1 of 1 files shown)


blade2005:
Can I get someone to review this. I could really use this.

PR #506, opened 2016-11-10
Add support for preserving empty fields

ianunruh:
Currently, when calling `DocType#to_dict`, any empty collections or `None` will be omitted from the resulting dictionary. The intention of this behavior is to prevent the case where accessing a field with `multi=True` actually changes it as a side-effect. However, for my use case, I wish to preserve the fact that there's an empty collection, as I recreate all documents from scratch at index time.

This change will allow the caller to override this behavior when calling `DocType#to_dict`. It could also be extended to `DocType#save` if desired, but I use the bulk helper so I have no need for this functionality in `save`.

If this change is not desired, there is a workaround, but it's pretty ugly and relies on accessing a private attribute on the document:

```python
dct = document.to_dict()
for key, value in document._d_.iteritems():
    if value in ([], {}, None):
        dct[key] = value
```

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -248,8 +248,8 @@ class DocType(ObjectBase):
             **doc_meta
         )
 
-    def to_dict(self, include_meta=False):
-        d = super(DocType, self).to_dict()
+    def to_dict(self, include_meta=False, include_empty=False):
+        d = super(DocType, self).to_dict(include_empty=include_empty)
         if not include_meta:
             return d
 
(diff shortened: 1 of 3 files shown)


HonzaKral:
@ianunruh even given your use case I don't understand why this behavior is desired. Could you please share how this behavior (preserving empty fields) will help your code? Thank you!

ianunruh:
@HonzaKral it turned out to be a bad design on my end, which is why I went ahead and closed this PR. Thanks for taking a look though!

PR #516, opened 2016-11-28
search's source value is lost when False or throws exception when list during clone

mumino:


elasticsearch_dsl/search.py
```diff
@@ -294,7 +295,10 @@ def _clone(self):
 
         s._response_class = self._response_class
         s._sort = self._sort[:]
-        s._source = self._source.copy() if self._source else None
+        if hasattr(self._source, "copy"):
```
HonzaKral: actually since `copy.copy` also works on `dict`s we don't need this `if`.

mumino: I left it if something needs special care. but I don't know any special case atm.

HonzaKral:
Thank you very much for this fix!

levylll:
I still get this error....
I install my elasticsearch_dsl with pip.
😢

levylll:
@HonzaKral Please upload the new package to pypi...

PR #517, opened 2016-11-29
Test on Elasticsearch 5.0.1 on TravisCI

brainix:


brainix:
...and Elasticsearch 5.0.2 just came out!  So test against that too.

mboynes:
It looks like ES is actually failing here since Travis uses Java 7 and ES 5.x requires Java 8 ([see line 173 in this job](https://travis-ci.org/elastic/elasticsearch-dsl-py/jobs/179874832#L173), for example), and the tests aren't using the installed version of ES at all. I didn't dig into it, but maybe that's why there are 47 skipped tests.

Issue #202, opened 2015-07-15
empty string values don't work


willkg:
I'm using master tip (423c4f20d5f28599145712bd11ac9ce23e616f9a).

If I define a `DocType` like this:

```
import elasticsearch_dsl as es_dsl
from elasticsearch_dsl import Doctype


class ProductDocType(DocType):
    id = es_dsl.Integer()
    name = es_dsl.String(index='not_analyzed')
```

Then empty string names should end up with an empty string token. I should be able to do a term filter with the empty string and get back products with no name.

However, it looks like `ObjectBase.to_dict` ignores empty values. So if I create an item with an empty value and save it, the empty value never makes it to Elasticsearch:

```
prod = ProductDocType(id=1, name='')
prod.save()
```

Thus when I later filter for `name=''`, I get nothing back:

```
>>> len(ProductDocType.search().filter('term', name='').execute())
0
```

https://github.com/elastic/elasticsearch-dsl-py/blob/423c4f20d5f28599145712bd11ac9ce23e616f9a/elasticsearch_dsl/utils.py#L440

What's the `not v` part for? Why reject falsey values?

HonzaKral:
Ah, that is good point.

I added this code after introducing the option to have blank fields to avoid the scenario where accessing a field causes the field to change value. Otherwise having:

    class MyDoc(DocType):
        tags = String(multi=True, index='not_analyzed')

and doing:

    md = MyDoc()
    md.tags

Would alter the value of the doc which would now be `{'tags': []}`.

I see from your example that it was not the correct solution. What do you think would be better? We can leave the empty values in there or only skip empty containers (`{}` and `[]`).

Thanks for the report!

willkg:
So:

* values that we don't want to send: `[]`, `{}`, `()`, `set()`.
* values we do want to send: `0`, `''`

Is `None` valid? Are there other valid/invalid values that are falsey?

willkg:
Thank you! :)

PR #529, opened 2016-12-07
Allows doc values to be empty strings in faceted search

AgentJeremyWelkley:
After #202 this check should have been updated to allow empty strings.


HonzaKral:
Thank you for the patch!

PR #530, opened 2016-12-07
added django-like exclude method to search request

AgentJeremyWelkley:


HonzaKral:
Thanks for the patch!

PR #536, opened 2016-12-17
Remove deprecated String field from docs

mattjmcnaughton:
Fix #534 

Version 5.0 deprecated the `String` field in favor of `Text` (if analyzed) and
`Keyword` (if not analyzed). This commit updates the remainder of the
documentation to reflect the new fields.

Signed-off-by: mattjmcnaughton <mattjmcnaughton@gmail.com>

PR #538, opened 2016-12-17
Remove deprecated String from field.py

### LINKED ISSUE #537: Remove `String` from `field.py`
My understanding is that `String` has been deprecated in favor of `Text` and `Keyword` in 5.0.0. However, the `String` class still exists in `elasticsearch_dsl/field.py`. As best I can tell, no other part of the code uses it, and [this pr](https://github.com/elastic/elasticsearch-dsl-py/pull/536) removes all mentions from the documentation.

I'll attach a pr in case its helpful. It's also entirely possible that I'm missing a reason why its important to keep it around, in which case feel free to close my pr.

mattjmcnaughton:
Fix #537 

As all use of `String` has been deprecated, we no longer need the class
in `field.py`. This commit removes the class definition.

Relates to #536. I figured since all of the code/documentation references to `String` are gone, its safe to go ahead and delete it.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -223,14 +223,6 @@ class Date(Field):
         except Exception as e:
             raise ValidationException('Could not parse date from the value (%r)' % data, e)
 
-class String(Field):
-    _param_defs = {
-        'fields': {'type': 'field', 'hash': True},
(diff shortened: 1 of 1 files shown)


HonzaKral:
string is still valid type in elasticsearch so it should remain, to be removed in 6.0. Thanks for the patch!

PR #554, opened 2017-01-07
Add support for Binary fields

### LINKED ISSUE #539: Support for Binary fields?
Is support for Binary fields planned at any point, or somewhere I'm missing?

mattjmcnaughton:
Fix #539 

Add support for a binary type, which accepts a binary value as a Base64
encoded string.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -289,3 +289,6 @@ class GeoShape(Field):
 
 class Completion(Field):
     name = 'completion'
+
+class Binary(Field):
+    name = 'binary'


HonzaKral:
I think that this would require a little bit more to accept binary data - not just strings. I think the library should do the base64 encoding/decoding. The question is how to represent the data? Probably `bytes`?

mattjmcnaughton:
Hmm, to be honest, I'm still trying to wrap my head around this. The return values for the `b64encode` and `b64decode` functions from the [base64](https://docs.python.org/2/library/base64.html) are both of type `str` as best as I could understand. Which is why I thought it would be ok to always keep the data in the same format. But I certainly could be missing something.

Definitely happy to make any necessary changes!

mattjmcnaughton:
@HonzaKral Just pinging for your thoughts on this whenever you get the chance. No rush!

I'm happy to make whatever changes, but am a little confused about what the input to the `base64` encoding would look like.

amirma:
@mattjmcnaughton  What was the reason this change was never merged?

Issue #560, opened 2017-01-19
Bug - new query classes cannot be serialized


dorony:
For example, the DisMax class. The problem here is that it defines the field `_params_def` instead of `_param_defs`.

HonzaKral:
Thank you for the report and the fix, this has been fixed by #561

dorony:
@HonzaKral Hi, when do you intend to publish the fixed version to PyPi?

PR #561, opened 2017-01-19
Bug - new query classes cannot be serialized

dorony:
Fixes issue: https://github.com/elastic/elasticsearch-dsl-py/issues/560


HonzaKral:
Thank you for the fix and the report!

PR #571, opened 2017-01-30
fix error on aggregation over nested field

erebus1:
mapping: 
`
    "tabs": {

    "type": "text",
    "fields": {
        "keyword": {
            "type": "keyword"
        }
    }

}`

`
      q.aggs.bucket('test', 'terms', field='tabs.keyword')

      r = q.execute()

      q.aggregation.test.buckets
`
raises:
exceptions:TypeError: 'String' object has no attribute '__getitem__'

### DIFF
--- a/elasticsearch_dsl/mapping.py
+++ b/elasticsearch_dsl/mapping.py
@@ -47,6 +47,8 @@ class Mapping(object):
                 field = field[step]
             except KeyError:
                 return
+            except TypeError:
+                return field
         return field
 
     def _collect_analysis(self):
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import sys
 from os.path import join, dirname
 from setuptools import setup, find_packages
 
-VERSION = (5, 1, 0)
+VERSION = (5, 1, 1)
 __version__ = VERSION
 __versionstr__ = '.'.join(map(str, VERSION))
 


karmi:
Hi @erebus1, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/571.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

HonzaKral:
I am sorry, I am unable to replicate this issue. I added a test in 15c6723b0c6dc5da28f752bf6023392928c26c6e that contains the same scenario and it works fine so I will close this PR.

Please feel free to reopen with additional information.
Thank you!

erebus1:
Yes, previous example wasn't correct. 
The problem is when we make aggregation using searcher from DocType

Here is an example: (using mapping from prev post)
`
from elasticsearch_dsl.document import DocType

class MyIndex(DocType):
    tabs = String()

searcher = MyIndex.search()
searcher.aggs.bucket('tabs', 'terms', field='tabs.keyword')
r = searcher.execute()

r.aggregation.test.bucket   # raises TypeError: 'String' object has no attribute '__getitem__'
`

erebus1:
@karmi, I've added email in my profile

erebus1:
@HonzaKral , can you reopen this issue?

Issue #574, opened 2017-02-03
Bug: `FacetedSearch.sort` can only send 1 argument


wimglenn:
Commit [8dbe180](https://github.com/elastic/elasticsearch-dsl-py/commit/8dbe180efba75e4a4c3e465d626d8e8375883493) added `FacetedSearch.sort` method but it doesn't use the `Search.sort` argument signature correctly.  

When `FacetedSearch` calls `Search.sort`, it needs to unpack the argument.  Otherwise, the `FacetedSearch` class is restricted to only ordering by one field, even though the `Search` class has more flexibility.  

Allowing FacetedSearch class to sort by multiple fields is a backwards-breaking change (since now using a string here will be unpacked into the single characters!), but the [`FacetedSearch` is currently documented as an experimental API and subject to change](http://elasticsearch-dsl.readthedocs.io/en/latest/faceted_search.html) so I didn't think that is too much of a big deal.

PR #575, opened 2017-02-03
Improve handling of the FacetedSearch sort keyword-argument

wimglenn:
Closes https://github.com/elastic/elasticsearch-dsl-py/issues/574

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -30,3 +30,4 @@ Andy Barilla <andrew.barilla@one.verizon.com>
 Rajiv Bakulesh Shah <brainix@gmail.com>
 Gabor Nagy <mail@aigeruth.hu>
 Shezad Khan <shaz@revl.world>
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thanks for this change, it looks good.

Just one detail - could you make it backwards compatible? Even when this is the proper way to do sort there might be people already just passing in a single field and this change would break their code. perhaps an `if isinstance(self.sort, (list, tuple)):`?

wimglenn:
OK, I added the compat.  Can I recommend the type check should be removed if/when the api is finalized (see Zen of python no. 13)

HonzaKral:
Awesome, thank you very much!

> Can I recommend the type check should be removed if/when the api is finalized (see Zen of python no. 13)

Absolutely, I will document only the tuple method and will add a note to remove the `isinstance` check later.

PR #587, opened 2017-02-11
Update README with dev instructions

### LINKED ISSUE #586: Idiomatic way to run individual tests when developing
My understanding is that `python setup.py test` can be used to run the entire test suite.

But is there a preferred way to only run the tests from a single file or even a single test? I would assume that it would be `pytest test_file.py`, but if that's the case, I'm not sure how to install the `tests_require` dependencies from [setup.py](https://github.com/elastic/elasticsearch-dsl-py/blob/master/setup.py#L19), other than installing them manually through `pip`? Perhaps its something I'm doing funny with my virtual environment, but the test dependencies do not remain accessible after running `python setup.py test`.

After I learn it, I'm happy to add whatever the preferred method is to `README.rst` if that would be helpful.

Thanks for your help!

mattjmcnaughton:
Fix #586 

Add instructions to the README for downloading the development/test
dependencies, as well as instructions on interacting the test suite.

### DIFF
--- a/README
+++ b/README
@@ -216,6 +216,49 @@ existing ``dict``, modifying it using the API and serializing it back to a
     # Convert back to dict to plug back into existing code
     body = s.to_dict()
 
+Development
+-----------
+
+To install all of the dependencies necessary for development, run:
+
+.. code:: bash
+
+    $ pip install -e .[develop]
+
+To run all of the tests for ``elasticsearch-dsl-py``, run:
+
+.. code:: bash
+
+    $ python setup.py test
+
+Alternatively, it is possible to use ``test_elasticsearch_dsl/run_tests.py``,
+which wraps `pytest <http://doc.pytest.org/en/latest/>`_,
+to run subsets of the test suite. Some examples can be seen below:
+
+.. code:: bash
+
+    # Run all of the tests in `test_elasticsearch_dsl/test_analysis.py`
+    $ python -m test_elasticsearch_dsl.run_tests test_elasticsearch_dsl/test_analysis.py
+
+    # Run only the `test_analyzer_serializes_as_name` test.
+    $ python -m test_elasticsearch_dsl.run_tests test_elasticsearch_dsl/test_analysis.py::test_analyzer_serializes_as_name
+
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -216,6 +216,37 @@ existing ``dict``, modifying it using the API and serializing it back to a
...
+
+    $ python setup.py test
+
+Alternatively, it is possible to use ``pytest`` to run subsets of the test
+suite. Some examples can be seen below:
+
+.. code::bash
+
+    # Run all of the tests in `test_elasticsearch_dsl/test_analysis.py`
+    $ pytest test_elasticsearch_dsl/test_analysis.py
```
HonzaKral: we have a `run_tests.py` wrapper in the `test_elasticsearch_dsl` directory, it would be better to point to that and mention that it is a thin wrapper around `pytest` - that way we are fw compatible in case we need to add additional setup commands in the future.

README
```diff
@@ -216,6 +216,37 @@ existing ``dict``, modifying it using the API and serializing it back to a
...
+
+    # Run only the `test_analyzer_serializes_as_name` test.
+    $ pytest test_elasticsearch_dsl/test_analysis.py::test_analyzer_serializes_as_name
+
+If trying to run a test from `test_elasticsearch_dsl/test_integration/`,
+ensure that you have a test instance of ElasticSearch running for which a
+test connection can be created. Integration tests will be skipped if it is
+not possible to create an ElasticSearch connection.
```
HonzaKral: It would be good to mention that the luster should be empty and destructive changes will occur, a big warning is probably in place.

Also pointing to the `TEST_ES_SERVER` that can be used to point to different hostname/port might be useful - https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/helpers/test.py#L12-L31

README
mattjmcnaughton: Great idea! Will do!

README
mattjmcnaughton: That is a great idea - definitely don't want people running this by accident without realizing.

mattjmcnaughton:
Thanks for the feedback! I updated my commit with your requested changes.

PR #593, opened 2017-02-18
Add virtualenv to setup in Dev environment

trishnaguha:
This PR contains info about creating and activating virtualenv to get started with development environment setup.

### DIFF
--- a/README
+++ b/README
@@ -219,6 +219,13 @@ existing ``dict``, modifying it using the API and serializing it back to a
 Development
 -----------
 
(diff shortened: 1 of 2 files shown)


README
```diff
@@ -219,6 +219,13 @@ existing ``dict``, modifying it using the API and serializing it back to a
 Development
 -----------
 
+Activate Virtual Environment:
```
HonzaKral: Could we also include a link to what virtual env is? Probably just point to http://docs.python-guide.org/en/latest/dev/virtualenvs/ or some other doc that describes what it is.

PR #594, opened 2017-02-18
Fixed command to run tests in README.

RatanShreshtha:


karmi:
Hi @RatanShreshtha, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/594.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

HonzaKral:
Thanks for the patch!

PR #596, opened 2017-02-18
Add test and doc for Search.exclude() in PR #530

nonbeing:
Added `test_exclude()` in `test_elasticsearch_dsl/test_search.py` for `Search.exclude()` added in PR #530.
Updated `docs/search_dsl.rst` with a simple example of `Search.exclude()`
Updated the `pip install` in the README to work for all shells.

### DIFF
--- a/README
+++ b/README
@@ -223,7 +223,7 @@ To install all of the dependencies necessary for development, run:
 
 .. code:: bash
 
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thanks for the patch!

PR #597, opened 2017-02-19
Add method, test and doc for Search.delete()

nonbeing:
Added Search.delete() that simply delegates to `delete_by_query()` in the low-level `elasticsearch-py` client.
This obviates #351 since delete_by_query is directly available in the low-level client (no need for a bulk delete since ES 5.x).

Added corresponding test to `test_elasticsearch_dsl/test_search.py` and doc to `docs/search_dsl.rst`.

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -84,6 +84,19 @@ explicitly:
 
     print(s.to_dict())
 
+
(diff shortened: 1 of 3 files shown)


HonzaKral:
Thank you for the patch, @nonbeing !

PR #602, opened 2017-02-25
Updated docs/index.rst

RatanShreshtha:
In Compatibility section added lines to show elasticsearch-dsl-py is compatible with Elasticsearch 5.x


HonzaKral:
Thanks for the patch!

Issue #614, opened 2017-03-08
Add `half_float` field


vforgione:
I have a use case for the new `half_float` numeric field type. Currently I have a custom field set up for my uses, but in trying to automate some of my work using the `Mapping.from_es` function it breaks when the field is used in a document mapping.

Relevant docs from Elastic: https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html

PR #615, opened 2017-03-08
Added `half_float` field

vforgione:
Added an entry in _fields.py_ for the new `half_float` type.

Fixes https://github.com/elastic/elasticsearch-dsl-py/issues/614


HonzaKral:
Thank you for the patch!

PR #626, opened 2017-03-20
Allow Request.index() to receive a list or tuple

f-santos:
Allow index to be updated with a list like when creating the Search object.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -135,13 +135,24 @@ class Request(object):
 
             s = Search()
             s = s.index('twitter-2015.01.01', 'twitter-2015.01.02')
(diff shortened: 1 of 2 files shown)


HonzaKral:
Hi, when you create a `Search` object you can pass in a list even now, this PR breaks existing code.

Can you specify when this would be used? Why couldn't the users just specify `s.index(*my_list)` instead?

Thank you!

f-santos:
Hi, didn't intend to break the current input, overlooked the single string case. Fixed it and added some tests.
The idea is to allow the same type of input as in the `Search` object argument, looks easier to handle it in the library and standardize both methods for specifying the indexes.

Thanks.

fxdgear:
@HonzaKral ping

HonzaKral:
Thank you fo rthe patch @f-santos !

Issue #627, opened 2017-03-21
Documentation update of breaking change in Suggester example


freko247:
Issue:
[Example](https://github.com/elastic/elasticsearch-dsl-py/blame/master/docs/persistence.rst#L126) of using Completion suggester is deprecated and is breaking in Elasticsearch > 5.x. This is well documented in the [breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/breaking_50_suggester.html#_completion_suggester_is_document_oriented).

Suggested fix:
Remove the payloads argument.

PR #628, opened 2017-03-21
Payloads is deprecated.

freko247:
Fixes: #627

Payloads is no longer supported, see [breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/breaking_50_suggester.html#_completion_suggester_is_document_oriented).

> Previously, context and completion suggesters supported an index-time payloads option, which was used to store and return metadata with suggestions. Now metadata can be stored as part of the the same document as the suggestion for retrieval at query-time. The support for index-time payloads has been removed to avoid bloating the in-memory index with suggestion metadata.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -123,7 +123,7 @@ If you want to create a model-like wrapper around your documents, use the
 
     class Post(DocType):
         title = Text()
-        title_suggest = Completion(payloads=True)
+        title_suggest = Completion()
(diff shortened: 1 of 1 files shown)


freko247:
I have signed the CLA. Maybe the user name doesn't match, I entered my email address as my github user name (the same email address that is linked to my github profile).

HonzaKral:
Thanks for the patch!

PR #633, opened 2017-03-25
Fix typo

signalw:
variable name is `blogs` not `blog`


HonzaKral:
Thank you for the fix!

PR #643, opened 2017-03-31
docs: documentation on MoreLikeThis functionality

eamonnmag:


docs/search_dsl.rst
```diff
@@ -414,6 +413,30 @@ endpoint) you can do so via ``execute_suggest``:
...
+    from elasticsearch_dsl Search
+
+    my_text = 'I want to find something similar'
+
+    s = Search()
+    # We're going to match based only on two fields, in this case text and title
+    s = s.query(MoreLikeThis(like=my_text, fields=['text', 'title]))
+    # You can also exclude fields from the result to make the response quicker in the normal way
+    s = s.source(exclude=["text"])
+    response = s.execute().to_dict()
```
HonzaKral: Could we avoid working with the raw `dict`s here? I understand it is sometimes needed but we should document the full API which includes the `Hit` wrapper, maybe just add a tiny `for` loop that will just print out the `title` fields of found documents?
```python
for hit in response:
    print(hit.title)
```

eamonnmag: Sure!

venix: Hey @HonzaKral , sorry for coming out of the blue. I am executing this query on elasticsearch 5.5  but doesn't return any results, but returns when I do a norma multi_match search. Is this supported by all versions? 

Here is a snippet from my code: 
` 
  more_like_this = 'Something awesome'
  search = ProductDoc.search()
  query = MoreLikeThis(like=more_like_this, fields=['title', 'description', 'tags', 'categories'])
  search = search.query(query)
  return search.execute().to_dict()
`
Am I doing something wrong? Also tried the code above.

Also @eamonnmag  the second import looks wrong  maybe change it from
`from elasticsearch_dsl Search`
to 
`from elasticsearch_dsl import Search` ?

SalahAdDin:
:+1:

PR #646, opened 2017-04-04
create q queries dynamically

gladsonvm:
This code fragmented is from my own answer to this SO Post
http://stackoverflow.com/questions/43138089/elasticsearch-dsl-python-unpack-q-queries/

I think this may help other developers to create q queries dynamically. :)

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -192,6 +192,36 @@ to directly construct the combined query:
     )
     s = Search().query(q)
 
+You may  `unpack Q queries dynamically` as given below
+
+.. code:: python
+    q_queries = [
+              {
+                "operation": "or",
+                "queries":
+                  [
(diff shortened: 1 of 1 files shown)


HonzaKral:
Hi, I am curious what is the purpose of this documentation snippet? It adds another layer of abstraction on top of elasticsearch-dsl which is already an abstraction on top of the query dsl. It also hides away all, very necessary, details such as query type.

You don't need to pass in some intermediary representation - instead of creating your `q_queries` just create a `Query` object, you can then combine them using standard python mechanism - either collect them in a list to then pass into a `bool` query or combine them directly using the appropriate operation (`|`). There is no need for another abstraction.

I am afraid I cannot accept this pull request, it is too specific for your use case and would be just too confusing and lead people down the wrong path.

PR #647, opened 2017-04-07
Documentation for filter_path

jzbahrai:
There was no documentation for filter_path in the dsl. PR to fix that. (I don't know if you merge up versions) but it could go back all the way to 2.2.

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -9,7 +9,7 @@ The ``Search`` object represents the entire search request:
   * queries
 
   * filters
-
+  
   * aggregations
 
   * sort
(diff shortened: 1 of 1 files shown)


HonzaKral:
I am reluctant to merge this because this essentially renders the entire library very fragile - you wouldn't be able to rely on the response at all and other features might not work either - even the `to_dict` may not help

jzbahrai:
I don't think this is about using the library in a specific manner. It's more about using ES with the options it offers? At least that's what I am using filter_path for and it took me a while to figure out how to use it with the dsl as there wasn't docs.

HonzaKral:
Problem is that the example in the docs would actually result in the library being unusable and throwing weird errors while obscuring others - for example if you had unavailable shards you would never know.

Most time there is also a better solution like setting size to 0 or use source filtering to only return some fields.

In general we cannot advertise `filter_path`. It is fine if people familiar with elasticsearch use it, like you have, but they should be aware of all the shortcomings of that approach which would mean having intimate knowledge of the internals of `elasticsearch-dsl`. That defeats the ultimate goal of the library.

Thank you for the patch but unfortunately I cannot accept it.

jzbahrai:
thanks! I appreciate you taking the time to explain that.

PR #661, opened 2017-05-14
Minor logic fix in is_published method

maxxkrakoa:
The check that compares `datetime.now` with `published_from` should be inverted so that any values of `datetime.now` that are greater than or equal to `published_from` result in the `is_published` method returning True.


HonzaKral:
Thank you for the fix, could I please ask you to sign our CLA (0) so that I can merge it?

maxxkrakoa:
Done

HonzaKral:
Thank you for the quick reaction!

PR #662, opened 2017-05-14
Date Histogram can get int or datetime as key for lookup, lets try both.

fxdgear:
When the key for a datetimehistogram was a datetime, it would fail expecting the value to be an int.
Adding checker to check the object type and react accordingly.
Added test


HonzaKral:
Thanks and welcome to the team!

PR #674, opened 2017-06-10
Set key to 0 when it is None in DateHistogramFacet

### LINKED ISSUE #673: Error with DateHistogramFacet, bucket['key'] is None instead of int()
I am using Doctype with Date() field. When I want to use DateHistogramFacet I get this error: 
`TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

Error is raised in file faceted_search.py in DateHistogramFacet.get_value() function on line 144: `return datetime.utcfromtimestamp(int(bucket['key']) / 1000)`

`bucket` variable shows this values during error:

```
bucket = {Bucket} {'key_as_string': '1970-01-01T00:00:00.000Z', 'key': None, '...}
 doc_count = {int} 84
 key = {NoneType} None
 key_as_string = {str} '1970-01-01T00:00:00.000Z'
```
Before calling get_value() function `data` variable looks like this:
```
[{'key_as_string': '1970-01-01T00:00:00.000Z', 'key': None, 'doc_count': 4},
{'key_as_string': '1970-02-01T00:00:00.000Z', 'key': 2678400000, 'doc_count': 3},
{'key_as_string': '1970-03-01T00:00:00.000Z', 'key': 5097600000, 'doc_count': 0}, 
{'key_as_string': '1970-04-01T00:00:00.000Z', 'key': 7776000000, 'doc_count': 5}, 
{'key_as_string': '1970-05-01T00:00:00.000Z', 'key': 10368000000, 'doc_count': 17}, 
{'key_as_string': '1970-06-01T00:00:00.000Z', 'key': 13046400000, 'doc_count': 4}, 
{'key_as_string': '1970-07-01T00:00:00.000Z', 'key': 15638400000, 'doc_count': 8}, 
{'key_as_string': '1970-08-01T00:00:00.000Z', 'key': 18316800000, 'doc_count': 2}, 
{'key_as_string': '1970-09-01T00:00:00.000Z', 'key': 20995200000, 'doc_count': 1}, 
{'key_as_string': '1970-10-01T00:00:00.000Z', 'key': 23587200000, 'doc_count': 0}, 
{'key_as_string': '1970-11-01T00:00:00.000Z', 'key': 26265600000, 'doc_count': 6}, 
{'key_as_string': '1970-12-01T00:00:00.000Z', 'key': 28857600000, 'doc_count': 34}, 
.....]
```

nrsimha:
Fixes #673 

When using DateHistogramFacet and date is 1970-01-01 midnight Elasticseach is returning key as None. 

We need to set key to 0 as to convert key with datetime.utcfromtimestamp() we require it to be int type (not NoneType).


fxdgear:
Nice catch. Would you mind writing a Unit test to protect any regression in the future?

nrsimha:
Yes, sure.

To be able to test it we need to have item with date of 1970-01-01.

Can I change `created_at` date from 2014-03-03 to 1970-01-01 for record in `repos` test data?

fxdgear:
I was thinking you could write a simple unit test that just instantiates the Class. You then call the method both with and without a value for `bucket['key']`

nrsimha:
Added test.

HonzaKral:
Thank you for the fix @nrsimha !

PR #675, opened 2017-06-12
Error when getting result facets 

browniebroke:
I recently upgraded from v2 and I have an error while trying to access to facets attribute:
`AttributeError: 'FacetedResponse' object has no attribute 'facets'`

And when accessing again it doesn't crash anymore but returns an empty dictionary instead of the expected result I had with v2.

Trying to reproduce the error on a more minimal setup here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,22 +11,23 @@ before_install:
   - java -version
 
 python:
-  - "2.6"
-  - "2.7"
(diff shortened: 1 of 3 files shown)


browniebroke:
Going back to my environment as it seems unrelated to this library. Sorry for the trouble.

PR #682, opened 2017-06-26
Loading aggregation with metadata from dictionary

ohadravid:
Hi,

We got an error while using an aggregation with meta data.
This code:
```
agg_as_dict = {
        'terms': {'field': 'tags'},
        'aggs': {'max_score': {'max': {'field': 'score'}}},
        'meta': {'some': 'metadata'}
}
agg = A(agg_as_dict)
```
Fails with `ValueError: A() can only accept dict with an aggregation`.

I added a test and implemented a fix.

Thanks!
Ohad

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -3,6 +3,10 @@
 Changelog
 =========
 
+master
+------
+ * fix bug in loading an aggregation with meta data from dict
+
 5.3.0 (2017-05-18)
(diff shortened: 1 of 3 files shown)


HonzaKral:
Merged as 5bcb9ba84d72f46294728b5a66c6b02b0bc6dbb3

Thanks!

Issue #702, opened 2017-07-29
How to use boost_mode 


ijharulislam:
How to use "boost_mode": "replace" in DSL? Is it possible to use it with SF, if so, how to do it?

HonzaKral:
Just passing `boost_mode='replace'` to `.query()` or `Q()` should do thr trick:

```python
from elasticsearch_dsl import Search, SF
s = Search().query('function_score', boost_mode='replace', functions=[SF('random_score')])
```

ijharulislam:
Thanks. It is working.

PR #703, opened 2017-07-29
added boost_mode to function_score #702

bloogrox:
ability to set `boost_mode` in a function_score

example:

```
GET /_search
{
    "query": {
        "function_score": {
            "query": { "match_all": {} },
            "boost": "5",
            "random_score": {}, 
            "boost_mode":"multiply"
        }
    }
}
```

more:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#score-functions

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -162,6 +162,7 @@ class FunctionScore(Query):
         'query': {'type': 'query'},
         'filter': {'type': 'query'},
         'functions': {'type': 'score_function', 'multi': True},
+        'boost_mode': {},
     }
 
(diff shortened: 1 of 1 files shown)


bloogrox:
how to use it

```
s.query = dslq.FunctionScore(
    query=s.query,
    functions=[dslq.SF('random_score')],
    boost_mode="replace"
)
```

HonzaKral:
Thanks for the patch @bloogrox, but it is actually not required. You can always pass arbitrary parameters when instantiating a quer. The parameters only need to be defined in `_params_def` if they require typecasting...

PR #712, opened 2017-08-08
repr Response object for more details

mapleray:
Some results of aggregations like cardinality  may not contains hits, and print result will show `<Response: []>`, which make user think the result is empty, but in fact, `response.aggregations` contains results.


mapleray:
local tests passed but travis not......

HonzaKral:
That is an unrelated issue as `elasticsearch-py` master branch is now 6.0 compatible. Thanks for the patch!

Issue #544, opened 2016-12-27
DateField returns datetime instead of date


andreyrusanov:
We faced an issue that DateField returns `datetime` type instead of `date`. The reason is that it use
`dateutil.parser.parse` function, that returns `datetime` anyway. 
I would vote to add DateTimeField, which will treat input data like a `datetime`, and modify DateField which will return `date`. I guess it is obvious way. 

Of course these changes might cause problems with backwards compatibility, so for minor version update it might be not nice to introduce those kind of changes, Instead, as a "quick fix", we can add a flag to DateField that marks what should be returned - date to datetime(by default it will be datetime to keep same behavior), 

What do you think?  My colleague already made custom field to fix this issue in our project, so one of us could make a pull request if you will like general idea.

HonzaKral:
I wonder what is the use case of havinf `date` instead of `datetime`? Since `datetime` is a subclass of `date` I didn't see the need to differentiate. Elasticsearch also doesn't have a date-level precision field.

Thank you

EricCat:
I'm also suffering from the same issue.

From this line:
https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/field.py#L221
might be will it be implemented?

For example: 
* input: Even if it's date format `2017-01-08`  
* output: Date(format="strict_date") `2017-01-08 00:00:00` 

` 00:00:00` is not required

HonzaKral:
@EricCat I am still trying to understand why that is an issue? Even when I give it a date without time information everything works just the same.

The output of the field is a `datetime` object which is a subclass of `date` so it shouldn't matter.

EricCat:
@HonzaKral Thanks update.
Hmm, I agree with you. Maybe It shouldn't belong to an issue.
However, if it could support this feature, will be better.
We don't need to parse it to date when querying the relating document.
Btw, `elasticsearch-dsl-py` is great!

andreyrusanov:
@HonzaKral we faced an issue when we mapped results from ElasticSearch to some serializers. We need to return Date and not DateTime, and we either had to implement additional logic to convert data or make custom wrapper on top of DateField. 
I know ElasticSearch has only `date` type which handles everything, but why not ot add a bit more flexibility on Python side?  Maybe in scope of DateField like @EricCat suggested.

HonzaKral:
[...]

PR #724, opened 2017-08-25
Date and DateTime field error fixed

mazhar266:
date field is not datetime field
new field added named date field

**Please, update the test cases, or help me update it**


HonzaKral:
Thanks for the patch, for now it is, however, out of scope for this library. See #544 for details.

PR #735, opened 2017-09-15
bug: if the key is long type

kongkongruye:
if the key is long type , it is also right


HonzaKral:
Hi @kongkongruye, sorry for the delay! Could you please sign the CLA at https://www.elastic.co/contributor-agreement so that we can merge this PR?

also please just use `isinstance(data, six.integer_types)` so this would also work in python 3.

Thanks!

HonzaKral:
Closing this PR since CLA was not signed.

PR #752, opened 2017-10-16
Prevent datetimes from dropping milliseconds

### LINKED ISSUE #750: Datetimes dropping miliseconds
The problems causes dates like '1970-01-01T00:17:47.045Z' in ElasticSearch to end up as a datetime like
datetime.datetime(1970, 1, 1, 0, 17, 47)                                                                                        

My expectation is that this value should produce a datetime like
datetime.datetime(1970, 1, 1, 0, 17, 47, 45000)

https://github.com/elastic/elasticsearch-dsl-py/blob/269fef7fa12333f7622c3694df75a1b296d87ae2/elasticsearch_dsl/field.py#L221

This is caused by integer division on the line referenced above. Changing this line to divide by 1000.0 fixes the issue.

mattjmcnaughton:
Fix #750

Fix an issue inwhich dates in elasticsearch which have
milliseconds may lose these milliseconds when retrieved,
because of how we manipulate the timestamp.

### DIFF
--- a/elasticsearch_dsl/faceted_search.py
+++ b/elasticsearch_dsl/faceted_search.py
@@ -145,7 +145,8 @@ class DateHistogramFacet(Facet):
             # so we need to set key to 0 to avoid TypeError exception
             if bucket['key'] is None:
                 bucket['key'] = 0
-            return datetime.utcfromtimestamp(int(bucket['key']) / 1000)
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thanks for the patch, much appreciated!

PR #759, opened 2017-10-25
Fix #692: Can't create custom stemming filter

### LINKED ISSUE #692: Can't create custom stemming filter
The stemming filter, requires the use a property called `name`, however the the library is using this property.

(https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stemmer-tokenfilter.html)

This code 
```python
light_stemmer_ = token_filter(name="minimal_english",type= "stemmer")
```

will produce this:
```json
          "filter" : {
                "minimal_english" : {
                    "type" : "stemmer"
                }
            }
```

while I would like this
```json
          "filter" : {
                "light_stemmer_" : {
                    "type" : "stemmer",
                    "name" : "minimal_english"
                }
            }
```

I suggest either changing the name of the variable the user is using, or allowing a variable like `_name` to become `name` when serializing.

as a workaround I am changed this line https://github.com/elastic/elasticsearch-dsl-py/blob/29d28a012a5a3a930e66cee56178208f21cb5fdf/elasticsearch_dsl/analysis.py#L33 to only pop if the type is not stemming.

like this
```python
         if self._builtin_type is 'stemmer' and 'name' in d[self.name] :
            d['name'] = d[self.name]['name']

        d = d.pop(self.name)
```
and them in my code I do
```python
light_stemmer_ = token_filter("light_stemmer_", "stemmer")
light_stemmer_.name = "minimal_english"
```
but I know it is a hacky solution

knixeur:
A default positional parameter in CustomAnalysis is called name which
conflicts with a parameter needed by the stemmer type filter. The parameter
 is renamed to filter_name instead of name to avoid conflict.

### DIFF
--- a/elasticsearch_dsl/analysis.py
+++ b/elasticsearch_dsl/analysis.py
@@ -19,9 +19,9 @@ class AnalysisBase(object):
 
 class CustomAnalysis(object):
     name = 'custom'
-    def __init__(self, name, builtin_type='custom', **kwargs):
+    def __init__(self, filter_name, builtin_type='custom', **kwargs):
         self._builtin_type = builtin_type
-        self._name = name
+        self._name = filter_name
         super(CustomAnalysis, self).__init__(**kwargs)
 
     def to_dict(self):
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thank you for the PR, much appreciated!

PR #766, opened 2017-11-08
conftest.py fixes

3lnc:
- `@fixture` instead of `@yield_fixture` (which is deprecated now)
- fixed addfinalizer call to actually do the job


3lnc:
This is weird. For some reason, travis installs `pytest == 2.9.2` for py3.3. Which is super old.

Had run tests locally on 3.3.7, works as intended.

HonzaKral:
Merged as b74441d

Thanks!

PR #769, opened 2017-11-15
6.0 - fix to look for 'result' key in es response

gingerwizard:
6.0 uses a 'result' key with the response type vs 'created'. This obviously doesn't resolve issues for 6.0, but was enough for me to move forward given i have have single types per index.


HonzaKral:
Fixed in 4e8241c, thanks for the pointer @gingerwizard !

PR #775, opened 2017-11-22
Explicit `scaling_factor` arg in `__init__`

3lnc:
Rationale for this: ability to get exception before putting the mapping in actual ES. `scaling_factor` is mandatory anyway.


HonzaKral:
Thank you for the patch, @3lnc !

PR #777, opened 2017-11-23
Classifiers fixes

3lnc:
Added `pypy` build – as far as I remember, at some time it was building in travis. Correct me if its no longer valuable. Overall, dsl seems to work fine under `pypy`.

Added `setup.py` classifiers according to current versions being tested.


HonzaKral:
Thanks for the patch, @3lnc!

PR #790, opened 2017-12-08
[WIP] Implemented OtherNested (name TBD)

HonzaKral:
Proof of concept only for now, more code, tests, and docs forthcoming...

Implementing a better interface for `Nested` fields which would allow easier, and more consistent, mapping definitions like:
```python
from elasticsearch_dsl import InnerDoc, DocType, Text, Keyword, Nested
class Comment(InnerDoc):
    title = Text()                                                        
    tags = Keyword(multi=True)                                            
                                                                                
class BlogPost(DocType):                                          
    comments = Nested(Comment)   
```

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -25,9 +25,11 @@ The mapping definition follows a similar pattern to the query dsl:
     m.field('category', 'text', fields={'raw': Keyword()})
 
     # you can also create a field manually
-    comment = Nested()
-    comment.field('author', Text())
-    comment.field('created_at', Date())
+    comment = Nested(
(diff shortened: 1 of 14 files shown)


3lnc:
Am I missing it, or there's no support for handling `dynamic` yet?

As of interface, seems that `__init__` is good place for it, for reusability, at least

HonzaKral:
`dynamic` works as expected:
```python
class ID(InnerDoc):
    title = Text()
    class Meta:
        dynamic = MetaField(False)
class DT(DocType):
    ids = OtherNested(ID)
```

Issue #785, opened 2017-11-29
Fields serialization (Long, Integer, Float, Double, etc.)


wowkin2:
Have similar issue like in [Problem with coerce? #21053](https://github.com/elastic/elasticsearch/issues/21053) during import from CSV
> I'm trying to make es cast some strings to integers and floats, but they seem to stay in as strings.

If I have field Double field and put there String with correct number "3.14" it will save me String in that Double type index. When I try to get it in Python code I'll get string without deserialization.
Only Boolean fields works correct (because it has redefined [_deserialize method](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/field.py#L261))

Do you plan to add serialization for other fields or can I do that or this should be done on my application side?

okke-formsma:
I've looked into this a bit.

There are three options for numeric datatypes that influence serialization: 'coerce', 'default', and 'ignore_malformed' (and combinations of these options). https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html.

Should elasticsearch_dsl accept a string like 'nan' or 'inf', even though elasticsearch will fail on them? (https://github.com/elastic/elasticsearch-py/issues/486)

3lnc:
@okke-formsma 
I guess, big part of the problem is how `coerce` works in ES.

 - it does not coerce data in `_source`
 - its inconsistent, i.e. it raise error when trying to coerce malformed string to numeric, meanwhile silently losing data while truncating decimals to integers

And to be honest, I have no idea if its actually usable for someone in the way its implemented now.

HonzaKral:
The reason why I originally didn't do serialization of numerical types is twofold - first I don't believe it should be needed when `json` fully supports those data types - there should never be any need or desire to express numeric values as strings for example.

Second reason is performance - currently we only coerce fields that absolutely require it and even as such it is a significant performance hit when instantiating a lot of objects, no matter what we do we need to make it optional to only be used when explicitly asked for.

okke-formsma:
@3lnc, @HonzaKral, thanks for pitching in. To summarise your posts; elasticsearch_dsl should be transparent in its dealing with types - junk in is junk out.

Maybe we should not parse 'false' to `False` for a boolean field to stay consistent? (undoing #788)

HonzaKral:
@okke-formsma that is a very good point about consistency with `"false"`, thank you for bringing it up!

[...]

PR #791, opened 2017-12-09
Added (de)serialisation for integer, float, ip and binary types.

okke-formsma:
Replaced 'Attachment' type (which was removed in 2.0) with 'Binary', added binary (de)serialization.

Casting the IP type to a `ipaddress.ip_interface` which is the closest to the elasticsearch type.

It's debatable whether the `float()` and `int()` casts are helpful, because the _source field should contain a float/integer whenever the right type was input anyway.

Discussed in #785.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -272,6 +272,7 @@ class Keyword(Field):
 
 class Boolean(Field):
     name = 'boolean'
+    _coerce = True
 
     def _deserialize(self, data):
         if data is None:
@@ -287,9 +288,9 @@ class Boolean(Field):
             raise ValidationException("Value required for this field.")
         return data
 
-
 class Float(Field):
     name = 'float'
+    _coerce = True
 
     def _deserialize(self, data):
         if data is None:
@@ -305,19 +306,17 @@ class ScaledFloat(Float):
     def __init__(self, scaling_factor, *args, **kwargs):
         super(ScaledFloat, self).__init__(scaling_factor=scaling_factor, *args, **kwargs)
 
-
 class Double(Float):
     name = 'double'
 
-
 class Integer(Field):
     name = 'integer'
+    _coerce = True
 
     def _deserialize(self, data):
         if data is None:
             return None
         return int(data)
-
 
 class Byte(Integer):
     name = 'byte'
@@ -330,6 +329,7 @@ class Long(Integer):
 
 class Ip(Field):
     name = 'ip'
+    _coerce = True
 
     def _deserialize(self, data):
         if data is None:
@@ -345,6 +345,7 @@ class Ip(Field):
 
 class Binary(Field):
     name = 'binary'
+    _coerce = True
 
     def _deserialize(self, data):
         if data is None:
@@ -391,6 +392,3 @@ class TokenCount(Field):
 
 class Murmur3:
     name = 'murmur3'
-
-class Percolator:
-    name = 'percolator'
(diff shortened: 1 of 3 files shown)


okke-formsma:
The new deserialization funtions raise `ValueErrors` if deserialization fails.

Should the `Date` type raise `ValueError`s too instead of `ValidationException`s?

elasticsearch_dsl/field.py
```diff
@@ -282,39 +287,74 @@ def clean(self, data):
...
     name = 'ip'
 
-class Attachment(Field):
-    name = 'attachment'
+    def _deserialize(self, data):
+        if data is None:
+            return None
+
+        # the ipaddress library for pypy, python2.5 and 2.6 only accepts unicode.
+        return ipaddress.ip_interface(six.u(data))
```
HonzaKral: We should be using `ip_address` instead of `ip_interface` here since that's the only supported value by elasticsearch - the `ip` field doesn't support ranges or blocks.

elasticsearch_dsl/field.py
```diff
@@ -345,3 +385,12 @@ class DateRange(Field):
...
     name = 'join'
+
+class TokenCount(Field):
+    name = 'token_count'
+
+class Murmur3:
+    name = 'murmur3'
+
+class Percolator:
+    name = 'percolator'
```
HonzaKral: I wonder if we should deserialize the percolator query into a `Query` object, what do you think?

elasticsearch_dsl/field.py
```diff
@@ -282,39 +287,74 @@ def clean(self, data):
...
+        # the ipaddress library for pypy, python2.5 and 2.6 only accepts unicode.
+        return ipaddress.ip_interface(six.u(data))
+
+    def _serialize(self, data):
+        if data is None:
+            return None
+        return str(data)
+
+class Binary(Field):
+    name = 'binary'
```
HonzaKral: we need `_coerce = True` here, otherwise the value won't be deserialized properly

elasticsearch_dsl/field.py
```diff
@@ -282,39 +287,74 @@ def clean(self, data):
...
+    name = 'byte'
+
+class Short(Integer):
+    name = 'short'
+
+class Long(Integer):
     name = 'long'
 
 class Ip(Field):
     name = 'ip'
```
HonzaKral: we need `_coerce = True` here, otherwise the value won't be deserialized properly

elasticsearch_dsl/field.py
```diff
@@ -282,39 +287,74 @@ def clean(self, data):
             raise ValidationException("Value required for this field.")
         return data
 
+
 class Float(Field):
     name = 'float'
 
-class HalfFloat(Field):
+    def _deserialize(self, data):
```
HonzaKral: we need `_coerce = True` here, otherwise the value won't be deserialized properly

elasticsearch_dsl/field.py
okke-formsma: oh right, it's the search API which should use an `ip_interface`.

okke-formsma:
I've added the _coerce fields.  I agree the percolator could be nicely represented by a query object. 

I don't have elasticsearch installed on my current pc (new install) so if you could change the integration tests that'd be great.

HonzaKral:
absolutely, I will take it from here, thank you so much, @okke-formsma!

HonzaKral:
Merged as  21a620d, thanks again for the help!

Simple integration test added as 95b6119

PR #795, opened 2017-12-13
Use string_types for string check

### LINKED ISSUE #792: Search does not support unicode indices
In elasticsearch_dsl/search.py, the `index` function checks to see if the arguments are of type `str` before appending them to the list of search indices. On Python 2, this test fails for unicode strings.

Instead, we can use `six.string_types` for Python 2 and 3 compatibility.

This problem may also occur elsewhere.

To reproduce:

```
from elasticsearch_dsl.search import Search
s = Search()
s.index('twitter-2015.01.01')  # works
s.index(u'twitter-2015.01.01')  # does not work
```

recursiveMake:
Fixes #792

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -148,7 +148,7 @@ class Request(object):
         else:
             indexes = []
             for i in index:
-                if isinstance(i, str):
+                if isinstance(i, string_types):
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thank you for the fix, @recursiveMake, much appreciated!

Issue #810, opened 2018-01-09
Deserialization of inner documents is not working


gilbsgilbs:
Hi,
Inner documents don't seem to be deserialized properly in 6.0.1. This diff adds a test case to reproduce the issue:

```diff
diff --git a/test_elasticsearch_dsl/test_integration/test_document.py b/test_elasticsearch_dsl/test_integration/test_document.py
index 324f645..b4c9ded 100644
--- a/test_elasticsearch_dsl/test_integration/test_document.py
+++ b/test_elasticsearch_dsl/test_integration/test_document.py
@@ -45,12 +45,14 @@ class Commit(DocType):
 
 class Comment(InnerDoc):
     content = Text()
+    created_at = Date()
     author = Object(User)
     class Meta:
         dynamic = MetaField(False)
 
 class PullRequest(DocType):
     comments = Nested(Comment)
+    created_at = Date()
     class Meta:
         index = 'test-prs'
 
@@ -67,7 +69,13 @@ class SerializationDoc(DocType):
 @fixture
 def pull_request(write_client):
     PullRequest.init()
-    pr = PullRequest(_id=42, comments=[Comment(content='Hello World!', author=User(name='honzakral'))])
+    pr = PullRequest(_id=42,
+                     comments=[
+                         Comment(content='Hello World!',
+                                 author=User(name='honzakral'),
+                                 created_at='2018-01-09T10:17:03.021184'),
+                     ],
+                     created_at='2018-01-09T09:17:03.021184')
     pr.save(refresh=True)
     return pr
 
@@ -110,6 +118,17 @@ def test_nested_inner_hits_are_wrapped_properly(pull_request):
     comment = pr.meta.inner_hits.comments.hits[0]
     assert isinstance(comment, Comment)
 
+
+def test_nested_inner_hits_are_deserialized_properly(pull_request):
+    s = PullRequest.search().query('nested', inner_hits={}, path='comments',
+                                   query=Q('match', comments__content='hello'))
+
+    response = s.execute()
+    pr = response.hits[0]
+    assert isinstance(pr.created_at, datetime)  # this works OK
+    assert isinstance(pr.comments[0].created_at, datetime)  # This fails, inner doc is not deserialized
+
+
 def test_nested_top_hits_are_wrapped_properly(pull_request):
     s = PullRequest.search()
     s.aggs.bucket('comments', 'nested', path='comments').metric('hits', 'top_hits', size=1)
```

[...]

PR #811, opened 2018-01-09
Add a test for inner documents deserialization.

gilbsgilbs:
Reproduces #810

### DIFF
--- a/test_elasticsearch_dsl/test_integration/test_document.py
+++ b/test_elasticsearch_dsl/test_integration/test_document.py
@@ -45,12 +45,14 @@ class Commit(DocType):
 
 class Comment(InnerDoc):
     content = Text()
+    created_at = Date()
     author = Object(User)
     class Meta:
         dynamic = MetaField(False)
 
 class PullRequest(DocType):
     comments = Nested(Comment)
+    created_at = Date()
     class Meta:
         index = 'test-prs'
 
@@ -67,7 +69,13 @@ class SerializationDoc(DocType):
 @fixture
 def pull_request(write_client):
     PullRequest.init()
-    pr = PullRequest(_id=42, comments=[Comment(content='Hello World!', author=User(name='honzakral'))])
+    pr = PullRequest(_id=42,
+                     comments=[
+                         Comment(content='Hello World!',
+                                 author=User(name='honzakral'),
+                                 created_at=datetime(2018, 1, 9, 10, 17, 3, 21184)),
+                     ],
+                     created_at=datetime(2018, 1, 9, 9, 17, 3, 21184))
     pr.save(refresh=True)
     return pr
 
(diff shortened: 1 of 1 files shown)


test_elasticsearch_dsl/test_integration/test_document.py
```diff
@@ -67,7 +69,13 @@ class Meta:
 @fixture
 def pull_request(write_client):
     PullRequest.init()
-    pr = PullRequest(_id=42, comments=[Comment(content='Hello World!', author=User(name='honzakral'))])
+    pr = PullRequest(_id=42,
+                     comments=[
+                         Comment(content='Hello World!',
+                                 author=User(name='honzakral'),
+                                 created_at='2018-01-09T10:17:03.021184'),
```
HonzaKral: we are now doing deserialization only when loading data form elasticsearch. That means you should be assiging a `datetime` object here. Could you please change this to `datetime(2018, 1, 9, 10, 17, 3, 21184)`?

Thank you!

HonzaKral:
Thank you very much the PR, I left on small comment but I can take care of it myself if you are busy.

gilbsgilbs:
@HonzaKral As long as we can fix and release this fast and clean, I highly prefer spending my time here rather than implementing a dirty workaround on my side.

test_elasticsearch_dsl/test_integration/test_document.py
gilbsgilbs: I'm not sure I have completely understood this point after all. Elasticsearch accepts strings as date types. It is `elasticsearch-py` that converts it to a string behind the hood. I'm totally fine with that. But I get confused with the IP type. If I set `comment.ip = '127.0.0.1'` and retrieve this from Elasticsearch, I'll get `comment.ip == ip_address.IPv4Address('127.0.0.1')` when retrieving the document. However, if I try to set `comment.ip = ip_address.IPv4Address('127.0.0.1')`, I'll get a serialization error. Basically, it means that I can't do something like `comment0.ip = comment1.ip`. Is it another issue? Or am I misunderstanding something?

HonzaKral: Yes, elasticsearch understand strings and you code actually worked without the comment I made, it just wasn't best practice (and people often look into the tests to see those).

I am getting no error when I have an `Ip` field defined on my `DocType` and then assigning `IPv4Address` I can correctly serialize the document (I added a test for it in 7e7c00838e68532522dd386045758de8f839640b). If that doesn't work for you it's definitely a bug and an issue would be great!

gilbsgilbs: Uuuuh my bad. Indeed, when I test locally it works. Still, I can reproduce the issue in my app. I have to figure out what causes `to_dict` to not serialize the `IPv4Address`.

gilbsgilbs: Got it. The IP field is in an inner object that I set "the old way": using a dictionary and not a proper object (e.g. `PullRequest(comment={'ip': ip_address('127.0.0.1')}])` instead of `PullRequest(comment=Comment(ip=ip_address('127.0.0.1')))`). Don't know if this is supposed to still be supported, but it's way way simpler in my case.

PR #812, opened 2018-01-09
Explicit signatures for few field types

3lnc:
This should provide better inspection for IDEs/IPython, etc.

Not sure on docstring format tho. I'm happy to document some/all existing fields/functions, but there's not too much of existing doc in codebase. I propose to use [napoleon/google docstrings](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). This approach seem to works quite well for real-world docs with params clarification/examples.

Pls share your thoughts on this

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -48,6 +48,10 @@ class Field(DslBase):
     _coerce = False
 
     def __init__(self, multi=False, required=False, *args, **kwargs):
+        """
+        :param bool multi: specifies whether field can contain array of values
+        :param bool required: specifies whether field is required
+        """
         self._multi = multi
         self._required = required
         super(Field, self).__init__(*args, **kwargs)
@@ -115,9 +119,18 @@ class Object(Field):
     _coerce = True
 
     def __init__(self, doc_class=None, dynamic=None, properties=None, **kwargs):
+        """
+        :param document.InnerDoc doc_class: base doc class that handles mapping.
+            If no `doc_class` is provided, new instance of `InnerDoc` will be created,
+            populated with `properties` and used
+        :param dynamic: whether new properties may be created dynamically.
+            Valid values are `True`, `False`, `'strict'`.
+            See https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html
+            for more details
+        :param dict properties: used to construct underlying mapping if no `doc_class` is provided
+        """
         if properties is None:
             properties = {}
-        self._doc_class = doc_class
         if doc_class is None:
             # FIXME import
             from .document import InnerDoc
@@ -127,6 +140,8 @@ class Object(Field):
                 self._doc_class._doc_type.mapping.field(name, field)
             if dynamic:
                 self._doc_class._doc_type.mapping.meta('dynamic', dynamic)
+        else:
+            self._doc_class = doc_class
 
         self._mapping = self._doc_class._doc_type.mapping
         super(Object, self).__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


3lnc:
ping @HonzaKral :)

elasticsearch_dsl/field.py
```diff
@@ -114,17 +118,30 @@ class Object(Field):
...
             # FIXME import
             from .document import InnerDoc
             # no InnerDoc subclass, creating one instead...
             self._doc_class = type('InnerDoc', (InnerDoc, ), {})
-            for name, field in iteritems(kwargs.pop('properties', {})):
+            for name, field in iteritems(properties):
                 self._doc_class._doc_type.mapping.field(name, field)
-            if 'dynamic' in kwargs:
-                self._doc_class._doc_type.mapping.meta('dynamic', kwargs.pop('dynamic'))
+            if dynamic:
```
HonzaKral: I believe this needs to be `if dynamic is not None` as `False` is a valid value that we have to propagate through.

elasticsearch_dsl/field.py
```diff
@@ -114,17 +118,30 @@ class Object(Field):
...
+            If no `doc_class` is provided, new instance of `InnerDoc` will be created,
+            populated with `properties` and used
+        :param dynamic: whether new properties may be created dynamically.
+            Valid values are `True`, `False`, `'strict'`.
+            See https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html
+            for more details
+        :param dict properties: used to construct underlying mapping if no `doc_class` is provided
+        """
+        if properties is None:
+            properties = {}
```
HonzaKral: we only need this `if doc_class is None`

elasticsearch_dsl/field.py
```diff
@@ -114,17 +118,30 @@ class Object(Field):
     name = 'object'
     _coerce = True
 
-    def __init__(self, doc_class=None, **kwargs):
-        self._doc_class = doc_class
+    def __init__(self, doc_class=None, dynamic=None, properties=None, **kwargs):
+        """
+        :param document.InnerDoc doc_class: base doc class that handles mapping.
+            If no `doc_class` is provided, new instance of `InnerDoc` will be created,
+            populated with `properties` and used
```
HonzaKral: I feel like, when we are defining args we should also check them - providing `properties` and `doc_class` together should raise an exception as it is not a valid combination.

Previously we could have pretended that we don't know as we are just passing things through, but now we no longer have the excuse :)

elasticsearch_dsl/field.py
```diff
@@ -114,17 +118,30 @@ class Object(Field):
     name = 'object'
     _coerce = True
 
-    def __init__(self, doc_class=None, **kwargs):
-        self._doc_class = doc_class
+    def __init__(self, doc_class=None, dynamic=None, properties=None, **kwargs):
+        """
+        :param document.InnerDoc doc_class: base doc class that handles mapping.
```
HonzaKral: please use `:arg ...` instead of `:param ....` for consistency throughout the library

3lnc:
Makes sense, will add fixes & tests.

Regarding docstrings, I agree regarding consistency, but what is this format, tho? I thought that "classic" sphinx/rst uses `param`.

And one more note on docstrings, what's your opinion on google style docs? Is it valuable in case we'll have more verbose docstrings, with examples perhaps? Because for now, `es-py` has pretty verbose docs, and most of the rtfd is compiled of that, for `es-dsl-py` this is different.

HonzaKral:
I tried to add some API docs to `es-dsl-py` (0), but we could definitely do more. Most of the docstrings for `es-py` are autogenerated :)

I am 100% pro verbose docstrings with examples, the more the better! I think the best example of style is the `Search` class where we have done the most but I am always happy for feedback/suggestions.

Thank you!


0 - http://elasticsearch-dsl.readthedocs.io/en/latest/api.html

Issue #828, opened 2018-02-01
Update nested fields with null values


golankopi:
When I'm trying to update a nested field property to null it remove when serialized and not sent to the ES.

For example:
{'my_field': {'A': 1, 'B': 1, 'C': None}}

will be serialized to:
{'my_field': {'A': 1, 'B': 1}}

PS it used to work for version 5.4.0 (I'm now upgrading my system to 6.1.0 and experiencing this issue)

Thanks!

HonzaKral:
I will reiterate this here as well - I don't believe you should be storing `None` values in elasticsearch, it is much better to leave the field out, as happens by default.

That said I will have a look at the PR... Thanks!

golankopi:
Hi, thanks,
I will explain the use case- we have a document with a certain field, then we want to update it so that field will not be relevant anymore, without start dealing with scripted deletions.

Any way, it is supprted by elastic it self.

Thanks Again!

HonzaKral:
I am sorry but I cannot in good conscience approve this change, it is too invasive, can break some code and the benefits are only to a corner case that you describe which is not common - there is no significant difference in either es or python in how null fields and missing fields are treated.

If you wish to preserve the null values anyway I'd suggest you override the `to_dict` method on the `InnerDoc` subclass you are using.

PR #829, opened 2018-02-01
Fix Nested fields None values serialization

golankopi:
In continuation to #828, here is the solution using the skip_empty argument


HonzaKral:
I am sorry, but this change is way too invasive and would break code using `AttrDict` directly which is still supported. I cannot accept this proposal.

PR #834, opened 2018-02-06
Do not combine Bool queries with different minimum_should_match values

### LINKED ISSUE #819: Bool queries with different minimum_should_match values are combined incorrectly
Combining two queries with different minimum_should_match values seem to result in one of the minimum_should_match values getting lost.

Example:
```
>>> q1 = query.Q('bool', minimum_should_match=2, should=[query.Q('term', field='aa1'), query.Q('term', field='aa2'), query.Q('term', field='aa3'), query.Q('term', field='aa4')])
>>> q2 = query.Q('bool', minimum_should_match=3, should=[query.Q('term', field='bb1'), query.Q('term', field='bb2'), query.Q('term', field='bb3'), query.Q('term', field='bb4')])
>>> q1 | q2
Bool(minimum_should_match=2, should=[Term(field='aa1'), Term(field='aa2'), Term(field='aa3'), Term(field='aa4'), Term(field='bb1'), Term(field='bb2'), Term(field='bb3'), Term(field='bb4')])
```

I would expect instead for a new `Bool(q1, q2)` object to be created so that differing `minimum_should_match` values are preserved.

I noticed the "TODO" in the code at https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/query.py#L98, so I suspect maybe more intricate behavior is expected.  For example, ideally, `Bools` with equal `minimum_should_match` values should be combined.

Proposed fix and tests illustrating the issue at:
bigbear-jrib/elasticsearch-dsl-py@aedda6366b0f72101969c3f71059e0e498d86c03

bigbear-jrib:
Closes #819

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -94,11 +94,10 @@ class Bool(Query):
 
     def __or__(self, other):
         for q in (self, other):
-            if isinstance(q, Bool) and not any((q.must, q.must_not, q.filter)):
-                # TODO: take minimum_should_match into account
+            if isinstance(q, Bool) and not any((q.must, q.must_not, q.filter, getattr(q, 'minimum_should_match', None))):
                 other = self if q is other else other
                 q = q._clone()
-                if isinstance(other, Bool) and not any((other.must, other.must_not, other.filter)):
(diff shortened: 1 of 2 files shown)


HonzaKral:
Merged as 88a9f96, thank you!

PR #835, opened 2018-02-07
Extend docsctring on FacetedSearch.search

konstin:
I found this method very confusing when I got started with FacetedSearch


HonzaKral:
Merged as 6576bfa18b7e774244f7a76ad2e38e1a8c097008, thank you for the patch!

PR #840, opened 2018-02-12
Define index from Document

### LINKED ISSUE #800: Allow index settings in Meta.index
Currently, when defining a new `DocType` you have to create a separate `Index` object if you wish to configure any index-level settings. With the deprecation of multiple types in elasticsearch 6.0 it might make more sense to just allow the index settings to configurable on the `DocType` itself in some fashion.

This would then allow even for situations where given `DocType` is present in multiple indices (time-based indices for example) meaning its `Meta.index` would specify a wildcard (`products-*` for example) and provide settings which would then allow it to generate and save an `IndeTemplate` and/or create any particular index via the `init` classmethod.

Any ideas and feedback is more than welcome!

Inspired by @3lnc's comment: https://github.com/elastic/elasticsearch-dsl-py/issues/779#issuecomment-350047587

HonzaKral:
Fixes #800 
TODO:
- [x] add `settings`, `analyzer`, `aliases`, etc support to `Meta`
- [x] megre `Index` and `IndexTemplate` - make `IndexTemplate` just a method on `Index` (`save_as_template(name, using=None)`)
- [x] add support for dynamic `using` param to all `Index` methods
- [x] clean up referential logic


docs/index.rst
```diff
@@ -203,7 +206,7 @@ You can see more in the :ref:`persistence` chapter.
 Pre-built Faceted Search
 ------------------------
 
-If you have your ``DocType``\ s defined you can very easily create a faceted
+If you have your ``Doccument``\ s defined you can very easily create a faceted
```
dsanders11: @HonzaKral, typo here, Doccument => Document

HonzaKral: thanks, fixed in 8ecf484

PR #679, opened 2017-06-20
add normalizer property for Keyword fields

billykern:
attempts to close: https://github.com/elastic/elasticsearch-dsl-py/issues/669

PR #847, opened 2018-02-24
import normalizer shortcut

billykern:
think this should've been included with: https://github.com/elastic/elasticsearch-dsl-py/pull/679


HonzaKral:
This has been fixed via https://github.com/elastic/elasticsearch-dsl-py/pull/856

Thanks for your patch!

PR #865, opened 2018-03-28
[WIP] Adding a first examples demonstrating complex data model and migrations

HonzaKral:


HonzaKral:
@3lnc thanks for the feedback!

problem with `author = User()` is that we don't know whether it should be `Object` or `Nested` which is very different. Also `User` is a class wrapping all users, you should be using it by saying `my_doc.author = User(username='honzakral', ...)` etc which wouldn't be possible (or straightforward) if we (also) used the class to designate the field.

fyi you can always say `Object(properties={'title': Text(), ...})` if you don't care about the `InnerDoc` subclass to use, though that wouldn't be recommended.

Does that make sense?

3lnc:
Just want to be totally sure that you saw `class User(Object):` in code above )

I mean, from interface point of view, there's clear way to specify if `User` is object or nested - in advance.

All complications to implement new interface is clear, especially when we're keeping old interface alive. But old interface with `Object(properties={...` — for me its hell. And that's exactly what I want to change sometime, even for instantiating single field in single model.

All pros with flexing `InnerDoc` into object or nested in different fields/models is clear and cool. Even having this alone — I'm all in for merging and having this in pypi version.

HonzaKral:
> Just want to be totally sure that you saw class User(Object): in code above )

thank you, I completely missed that, my apologies! That could indeed be nice if you don't care about the `InnerDoc` class but then how would assigning to that field look like? Right now the proper way is to assign instance of `User` (which is also a place where you can define your methods, validation, etc) but this would leave you working with raw dicts which is not desireable I think...

> All pros with flexing InnerDoc into object or nested in different fields/models is clear and cool. Even having this alone — I'm all in for merging and having this in pypi version.

That code should work in the released version, it is already there :)

3lnc:
Good question about assigning.

Errr. Rough stuff about assigning. `post.author = User(name='John')` is intuitive and not working :)

Need to make some prototypes and think a little bit more, I think. Django ORM is what I like in terms of interface, but implementing objects as attributes without relations in db – need to make sure if it actually makes sense.

HonzaKral:
@3lnc fwiw I think we need to give priority to how the field is used than to how it's defined since you only define it one time and use it (hopefully ;) ) all over your code.

Django ORM is great, but it is flat, they don't have to deal with nasty bits like (lists of) objects... :)

There are already some samples of this API in tests (grep for `Nested`, `Object`, or `InnerDoc`) and it didn't feel *too* unwieldy to write but of course any improvements would be more than welcome!

I will also probably add `add_comment` method or something similar to the example so it's clearer how it can be used...

examples/parent_child.py
```diff
@@ -0,0 +1,168 @@
...
+    number_of_replicas=0
+)
+# add the DocType classes to the index, their mappings will be merged
+BASE_INDEX.doc_type(Answer)
+BASE_INDEX.doc_type(Question)
+
+
+def setup():
+    " Create an IndexTemplate and save it into elasticsearch. "
+    index_template = BASE_INDEX.as_template('base')
```
amih90: `as_template` exists?

HonzaKral: ooops, it exists in `meta-index` branch, part of #840 

I got my branches crossed, sorry about that

examples/parent_child.py
```diff
@@ -0,0 +1,252 @@
...
+
+    class Meta:
+        def matches(hit):
+            " Use Answer class for child documents with child name 'answer' "
+            return isinstance(hit['_source']['question_answer'], dict) and hit['_source']['question_answer'].get('name') == 'answer'
+
+
+# construct basic index and populate with default settings. This will then be
+# used to create a template
+BASE_INDEX = Index('*.stackexchange.org')
```
amih90: Creating Index with "*" fails

"elasticsearch.exceptions.RequestError: TransportError(400, 'invalid_index_name_exception', 'Invalid index name [*.stackexchange.org], must not contain the following characters [ , ", *, \\, <, |, ,, >, /, ?]')"

examples/parent_child.py
```diff
@@ -97,28 +101,39 @@ def add_answer(self, user, content, created=None, accepted=False):
...
         s = s.params(routing=self.meta.id)
         return s
 
+    def save(self, **kwargs):
+        self.question_answer = 'question'
+        return super(Question, self).save(**kwargs)
+
     class Meta:
-        def matches(self, hit):
+        def matches(hit):
```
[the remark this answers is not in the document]
HonzaKral: done, thanks for pointing that out!

examples/parent_child.py
HonzaKral: This index should never be created directly. Always create an index returned by `get_index` function as the example shows

examples/parent_child.py
```diff
@@ -0,0 +1,254 @@
...
+    title='How do I use elasticsearch from Python?',
+    body='''
+    I want to use elasticsearch, how do I do it from Python?
+    ''',
+)
+q1.save()
+
+a1 = q1.add_answer(honza, "Just use `elasticsearch-py`!")
+
+# refresj the index so we can search right away
```
amih90: refresh typo

examples/parent_child.py
```diff
@@ -0,0 +1,254 @@
...
+    number_of_shards=1,
+    number_of_replicas=0
+)
+# add the DocType classes to the index, their mappings will be merged
+BASE_INDEX.doc_type(Answer)
+BASE_INDEX.doc_type(Question)
+
+
+def setup():
+    " Create an IndexTemplate and save it into elasticsearch. "
```
amih90: how can I add mapping of the classes to the index? should I call `init()`? for each class?

HonzaKral:
closing this PR as the examples are now part of #840

PR #867, opened 2018-03-28
A tiny initial percolation's implenetation.

SalahAdDin:
Parameters full list here:  https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-percolate-query.html#_parameters_10

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -211,6 +211,9 @@ class Indices(Query):
     name = 'indices'
     _param_defs = {'query': {'type': 'query'}, 'no_match_query': {'type': 'query'}}
 
+class Percolate(Query):
+    name = 'percolate'
+    _param_defs = {'percolate': {'field': 'query', 'document': {'field': 'query'}}}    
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thank you for the PR, @SalahAdDin. The `_params_def` in the class definition serve a different purpose, however, they don't map all the possible parameters in the body but instead only those that are of some known type like `query`. In this case the `_params_def` should not be defined on the class and everything should work fine. You'd then use the class as follows:
```python
raw_query ={
        "percolate" : {
            "field" : "query",
            "document" : {
                "message" : "A new bonsai tree in the office"
            }
        }
    }

query = Q('percolate', field='query', document={'message': 'A new bonsai tree in the office'})
```

if you'd remove the `_params_def` I'd be happy to merge the patch.

Thanks again!

SalahAdDin:
@HonzaKral  It's done.

HonzaKral:
Thank you, @SalahAdDin !

SalahAdDin:
Your welcome.

PR #879, opened 2018-04-12
Use string_types for string check (#795)

gleb-jiji:
This fix for #792 from #795 hasn't been merged into the 5.x branch but is necessary there as well.


HonzaKral:
thanks for the patch!

PR #888, opened 2018-04-23
Populate document.save/update result value

amih90:
I found very useful the return value of the persistence methods, but they are populated in a none standard way (with a boolean) in case of `save` and doesn't populated at all when update.


amih90:
@karmi added. is that okay now?

karmi:
@amih90, great, the check is green now!

amih90:
@HonzaKral Please consider merging it into the next release

HonzaKral:
@amih90 I am planning to merge this for `7.0` since it is a backwards incompatible change, thank you for the patch, I think it adds clarity to the confusion that was the boolean value :)

HonzaKral:
Merged as e3d4336

Thank you for the patch!

Issue #895, opened 2018-05-09
Document mapping with nested property does not respect `include_in_parent` parameter


dbmikus:
In the course of migrating to ES v6 from ES v2, I discovered a bug when defining a `DocType` subclass for persistence. If it has a nested field with `include_in_parent` set, that `include_in_parent` parameter will not be sent to the Elasticsearch cluster.

I've tested that this issue is present on the master branch. It is not present on the 2.x branch.

See this example code:

```
from datetime import datetime
import elasticsearch_dsl as esdsl

# Define a default Elasticsearch client
esdsl.connections.connections.create_connection(hosts=['localhost:9200'])

class Article(esdsl.DocType):
    offices = esdsl.Nested(
        include_in_parent=True,
        properties={
            'oid': esdsl.Integer()
        }
    )

    class Meta:
        index = 'blog'

    def save(self, **kwargs):
        return super(Article, self).save(**kwargs)

# create the mappings in elasticsearch
Article.init()

# create and save and article
article = Article(meta={'id': 42}, offices=[{'oid': 4}])
article.save()

article = Article.get(id=42)
print(article.offices)

# Display cluster health
print(esdsl.connections.connections.get_connection().cluster.health())
```

I intercepted the TCP packets sent to the cluster to verify that it did not contain `include_in_parent` and I did an `HTTP GET localhost:9200/blog` to confirm that it was not present in the mapping returned.

HonzaKral:
fixed by 2fd76081633fded8e2713c3959ded87ed940af94

PR #896, opened 2018-05-09
Allow extra field parameters to be serialized in JSON mapping

dbmikus:
In some cases, a field definition has extra parameters. For example, the
"nested" field type can have the extra parameter "include_in_parent". We
want to include these extra parameters when serializing the objects to
the JSON object we send to Elasticsearch.

This adds a class with a nested field with a couple field parameters set
and then makes sure that the option appears in the dictionary
serialization of that class's mapping.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -169,7 +169,7 @@ class Object(Field):
     def to_dict(self):
         d = self._mapping.to_dict()
         _, d = d.popitem()
-        d["type"] = self.name
+        d.update(super(Object, self).to_dict())
         return d
 
     def _collect_fields(self):
--- a/test_elasticsearch_dsl/test_document.py
+++ b/test_elasticsearch_dsl/test_document.py
@@ -63,6 +63,13 @@ class SecretDoc(document.DocType):
 class NestedSecret(document.DocType):
     secrets = field.Nested(SecretDoc)
 
+class NestedWithParams(document.DocType):
+    my_nested = field.Nested(
+        include_in_parent=True,
+        dynamic=False,
+        properties={'title': field.Text()})
+
+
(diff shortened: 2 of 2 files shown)


dbmikus:
This is for #895. Let me know if this is not the best way to solve the problem. It looks like all of the other `to_dict()` methods inside field.py end up calling the super class which handles putting the `_params` in the dictionary generated. I just added a simplified version of the same thing into the `fields.Object.to_dict()` method.

elasticsearch_dsl/field.py
```diff
@@ -170,6 +170,9 @@ def to_dict(self):
         d = self._mapping.to_dict()
         _, d = d.popitem()
         d["type"] = self.name
+        for k, v in iteritems(self._params):
+            if v not in ({}, []):
+                d[k] = v
```
HonzaKral: Thanks for the patch! I think this piece should be solved by calling `super().to_dict()` and merging it with `d` (which will actually make the previous line not needed).

Do you think you could update the patch ao i can merge it?

Thanks again!

dbmikus:
Made the fix, @HonzaKral.

HonzaKral:
Thanks @dbmikus, could I also ask you to sing our CLA (0) so we can merge it?

vinayan3:
I think I'm hitting this too. I set `enabled` to False on a [`Object`](https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html). 

I checked my mapping and it has all the fields being indexed. A fix for this soon would be awesome.

HonzaKral:
fixed in 2fd76081633fded8e2713c3959ded87ed940af94

dbmikus:
Thanks for getting this in there and sorry for being slow about the CLA process on my end.

PR #920, opened 2018-06-15
Fix Binary field in Python 3.

schuyler:
The `base64` module was changed in Python 3 to return `bytes` rather than `str`. This change caused elasticsearch-dsl to fail with a `TypeError` when trying to serialize Binary fields in Python 3. Adding a call to `.decode()` changes the return type back to `str`; in Python 2, this is a no-op.

Similarly, the default behavior of `Field.clean()` attempted to deserialize the Binary field data on `DocType.save()`, which naturally failed (with an 'incorrect padding' error) because the binary data wasn't base64-encoded yet. Since the Binary data is opaque, I assume it's acceptable to simply return the raw data from `.clean()`.

These two changes make it possible to correctly save and load documents containing Binary fields in Python 3.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -345,7 +345,7 @@ class Binary(Field):
     def _serialize(self, data):
         if data is None:
             return None
-        return base64.b64encode(data)
+        return base64.b64encode(data).decode()
 
 class GeoPoint(Field):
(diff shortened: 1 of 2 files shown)


schuyler:
Bump

HonzaKral:
Thank you for the PR and my apologies for it taking so long!

Issue #929, opened 2018-07-03
Document mapping doesn't accept any name but 'doc'


noamkush:
I tried to use the new 6.2 release and found that the following code crashes:
```
from elasticsearch_dsl import Document, Mapping

class MyDoc(Document):
    class Meta:
        mapping = Mapping('a')
```
With the following trace:
```
/lib/python3.6/site-packages/elasticsearch_dsl/document.py in __new__(cls, name, bases, attrs)
     30         new_cls = super(IndexMeta, cls).__new__(cls, name, bases, attrs)
     31         new_cls._index = cls.construct_index(index_opts, bases)
---> 32         new_cls._index.document(new_cls)
     33         return new_cls
     34

/lib/python3.6/site-packages/elasticsearch_dsl/index.py in document(self, document)
    121             raise IllegalOperation(
    122                 'Index object cannot have multiple types, %s already set, '
--> 123                 'trying to assign %s.' % (self._mapping.doc_type, name))
    124         self._doc_types.append(document)
    125         # TODO: do this at save time to allow Document to be modified after

IllegalOperation: Index object cannot have multiple types, doc already set, trying to assign a.
```
I tried to add `doc_type='a'` to the `Meta` class, but it didn't help. As far as I have tried, there's absolutely no way to have a mapping which is not named `'doc'`. This is also incompatible with existing indexes.

HonzaKral:
Thanks for the report, this is indeed a bug as we need to make sure we expose the `doc_type` option on `class Index`. I will provide a fix and release `6.2.1` asap.

jason-huling:
Hey @HonzaKral thank you for the fix. Do you have an example of how the doc_type should be specified in 6.2.1? I am using that version but get a similar `IllegalOperation` error when using the provided code:
```
class MyDoc(Document):
    class Meta:
        mapping = Mapping('a')
```
If I use the following code block, I get a slightly different error:
```
class MyDoc(Document):
    class Index:
        doc_type = "a"
```
With the following trace:
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File ".../lib/python3.6/site-packages/elasticsearch_dsl/document.py", line 32, in __new__
    new_cls._index.document(new_cls)
  File ...lib/python3.6/site-packages/elasticsearch_dsl/index.py", line 123, in document
    'trying to assign %s.' % (self._mapping.doc_type, name))
[...]

PR #930, opened 2018-07-03
Restores old DocType interface

3lnc:
DocType was refactored to Document, DocType as class was added to backaward compat, but its missing from high-level interface.
@HonzaKral  `from elasticsearch_dsl import DocType` is broken in latest PyPI release (6.2.0). Taking into account that this is minor one, I guess, this should be merged and re-uploaded to PyPI asap.

### DIFF
--- a/elasticsearch_dsl/__init__.py
+++ b/elasticsearch_dsl/__init__.py
@@ -3,7 +3,7 @@ from .aggs import A
 from .function import SF
 from .search import Search, MultiSearch
 from .field import *
-from .document import Document, MetaField, InnerDoc
+from .document import Document, DocType, MetaField, InnerDoc
 from .mapping import Mapping
 from .index import Index, IndexTemplate
 from .analysis import analyzer, char_filter, normalizer, token_filter, tokenizer


HonzaKral:
Thanks, good catch! I will add it to the planned `6.2.1` release with #929

3lnc:
Thanks. Any ETA on this?

HonzaKral:
`6.2.1` is out, thanks again for the quick patch!

GeorgeLubaretsi:
I think it makes sense to add `DocTypeMeta` compatibility as well

HonzaKral:
@GeorgeLubaretsi in what way? That was never really considered public api and is no longer around, what is the use case for it?

GeorgeLubaretsi:
`DocTypeMeta` is used in [django-elasticsearch-dsl](https://github.com/sabricot/django-elasticsearch-dsl/blob/v0.5.0/django_elasticsearch_dsl/documents.py#L53) for registering and mapping ORM models to ES documents.

On a second thought, I guess this should be resolved in `django-elasticsearch-dsl` if this was never a public API.

HonzaKral:
It should definitely be resolved there. I would be happy to help if there are some additional hooks or anything needed in `elasticsearch-dsl`, just open an issue or a PR. Thanks!

PR #938, opened 2018-07-07
Make 'Compatibility' in README more accurate

dsanders11:
I assumed `1.x` is no longer supported by master since it wasn't shown in the examples, if it's still supported than I can redo this PR to simply add the `5.x` branch mention.


HonzaKral:
Thank you for the clarification!

PR #939, opened 2018-07-07
Improvements to .travis.yml

dsanders11:
The current `.travis.yml` is testing using the `5.x` version of `elasticsearch-py` which isn't correct. Also bumped the ES version and added `6.3.x`


HonzaKral:
Thank you for the patch, very useful fixes!

PR #942, opened 2018-07-09
__json__ serializer.py

satels:
Use way https://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable


HonzaKral:
thank you for the patch, @satels. Do you have any use case in mind for this functionality?

Currently the serializer should never encounter any objects outside of this library which all implement `to_dict`. Not to mention that, if everything goes well (the pyhton code is in sync with elasticsearch mappings and only standard APIs are used), those should already be serialized to plain objects when they get to the serializer.

for serializing of custom data formats custom `Field` classes should be used instead.

HonzaKral:
Closing this since there is no traction and also no use case behind the PR

PR #976, opened 2018-08-02
Add `NestedFacet` to __all__

barseghyanartur:
Add `NestedFacet` to __all__


HonzaKral:
Thank you for the patch!

PR #988, opened 2018-08-15
Add  Facet to __all__

bfcrampton:
I would like to create my own custom subclass of `Facet` in my repo which requires this to be exposed.

### DIFF
--- a/elasticsearch_dsl/faceted_search.py
+++ b/elasticsearch_dsl/faceted_search.py
@@ -8,8 +8,8 @@ from .response import Response
 from .query import Terms, Nested, Range, MatchAll
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


HonzaKral:
The `Facet` class is not usable on its so we didn't expose it in `__all__`, but you can still import it:
```python
from elasticsearch_dsl.faceted_search import Facet
```
Does that work fo you or you think enough people would be interested in it so that we should add it to `__all__` so that people can d0 just:
```python
from elasticsearch_dsl import Facet
```

I am leaning towards just allowing the explicit import since it is a more advanced use case anyway...

bfcrampton:
Ah sorry I misunderstood how this worked, 
```
from elasticsearch_dsl.faceted_search import Facet
```
seems perfectly fine to me, thanks!

PR #994, opened 2018-08-21
document save with skip_empty parameter in to_dict

jvalhondo:
Document `save` method allows the parameter `skip_empty` to be set to `False` and passed to the method `to_dict`. 

Therefore the empty values (``None``, ``[]``, ``{}``) will be left on the document, and so indexed into the elasticsearch.

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -384,7 +384,7 @@ class Document(ObjectBase):
             if '_' + k in meta:
                 setattr(self.meta, k, meta['_' + k])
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thank you for the patch, could you please also include some tests for this functionality? Thank you!

jvalhondo:
@HonzaKral test included and passed. Think I implemented it at the right location with clear naming. Give me some feedback otherwise.
`test_elasticsearch_dsl/test_integration/test_document.py::test_save_without_skip_empty_will_include_empty_fields PASSED                                                                              [ 88%]`
Thank you!

PR #1000, opened 2018-08-26
Implementing hook for dynamically adding fields to mapping

safwanrahman:
For having the mapping dynamically generated, we need to have a hook that will get the fields dynamically. I believe, its needed by `django-elasticsearch-dsl` to generete Elasticsearch DSL mapping dynamically from Django model.
@HonzaKral As you have mentioned in your comment https://github.com/elastic/elasticsearch-dsl-py/pull/930#issuecomment-402189706, is it a correct approach to have a hook for `django-elasticsearch-dsl`? I am currently trying to upgrade it for `elasticsearch-dsl-py>6.1.0`

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -24,6 +24,7 @@ class DocumentMeta(type):
         attrs['_doc_type'] = DocumentOptions(name, bases, attrs)
         return super(DocumentMeta, cls).__new__(cls, name, bases, attrs)
 
+
 class IndexMeta(DocumentMeta):
     # global flag to guard us from associating an Index with the base Document
     # class, only user defined subclasses should have an _index attr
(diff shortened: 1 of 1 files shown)


safwanrahman:
Seems like it will not work! 😞

safwanrahman:
@HonzaKral Do you have any idea how to accomplish the feature to have a hook to dynamically generate the mapping?

HonzaKral:
There should be no need for an extra hook to add fields, you can always access the `Mapping` by doing `DocClass._doc_type.mapping` or `DocClass._index._mapping` and then you can easily call the `.field()` method there to add additional fields.

The only issue now is that if you add a field to the `DocClass._doc_type.mapping` it will not propagate up to the `Index` at `DocClass._index` (unless you create a new one) so `DocClass.init()` won't take it into account... I fixed that in ceac532835ba59e533c7d96250f85b25c85c5c79

Hope this helps!

PR #1033, opened 2018-10-12
Correct documentation of `source()`

dotJoel:
The `source()` had its arg incorrectly defined as `source` when it should have been `fields`.


HonzaKral:

PR #1043, opened 2018-10-26
Clarify creating a connection and its usages.

annawinkler:
I've been using this library and want to clarify creating a connection. I like using an alias for a single connection, so I added an example for that. I also updated some of the language for clarity.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -3,14 +3,14 @@
 Configuration
 =============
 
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thank you so much for this patch, I really appreciate any documentation patches and this is a great improvement in readability

annawinkler:
You're welcome! Thanks for merging this so quickly! Hopefully I can help out more as I use this library more 😃

PR #1046, opened 2018-10-29
Fix collections.abc deprecation warning in python 3.7.

annawinkler:
I'm using Python 3.7, and in my library I was getting a deprecation warning about the usage of collections in this library that I would like to fix. I think this library is supporting Python 2 and 3, so having the try / except will support the different Python version.

I did not see an existing issue for this so I'll create one.


HonzaKral:
Thank you for the catch and the patch!

Issue #1050, opened 2018-10-30
Implement update_by_query endpoint


emarcey:
I'm using the DSL for a project and I want to use the update_by_query endpoint. I didn't see any issues open regarding implementing it, so I've forked the repo and started development on including an UpdateByQuery object similar to Search.

HonzaKral:
Resolved in #1051, thanks for the great work!

emarcey:
Hey @HonzaKral, do you have an idea of when a version will be tagged with this fix in it?

HonzaKral:
I will try to get to it tomorrow or next week, stay tuned!

PR #1051, opened 2018-10-31
99/tc 499 implement update by query

emarcey:
Resolves issue #1050 

Creates object UpdateByQuery, with associated method in Index.

UpdateByQuery wraps the [_update_by_query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html) endpoint, which allows a search query & script to update a field in all matching docs.

Re-uses much of Search functionality (with fewer options, like no aggregations), and swaps Search's script_fields attribute with script that matches the scripting documented in the [elasticsearch API](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-using.html)

### DIFF
--- a/docs/update_by_query.rst
+++ b/docs/update_by_query.rst
@@ -1,0 +1,148 @@
+.. _update_by_query:
+
+Update By Query
+================
+
+The ``Update By Query`` object
+-------------------------------
+
+The ``Update By Query`` object enables the use of the 
+`_update_by_query <https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html>`_
+endpoint to perform an update on documents that match a search query.
+
+The object is implemented as a modification of the ``Search`` object, containing a 
+subset of its query methods, as well as a script method, which is used to make updates.
+
+The ``Update By Query`` object implements the following ``Search`` query types:
+  
+  * queries
+
+  * filters
+
+  * excludes
+
+For more information on queries, see the :ref:`search_dsl` chapter.
+
+Like the ``Search`` object, the API is designed to be chainable. This means that the ``Update By Query`` object
+is immutable: all changes to the object will result in a shallow copy being created which
+contains the changes. This means you can safely pass the ``Update By Query`` object to
+foreign code without fear of it modifying your objects as long as it sticks to
+the ``Update By Query`` object APIs.
+
+You can pass an instance of the low-level `elasticsearch client <https://elasticsearch-py.readthedocs.io/>`_ when
+instantiating the ``Update By Query`` object:
+
+.. code:: python
+
+    from elasticsearch import Elasticsearch
+    from elasticsearch_dsl import UpdateByQuery
+
+    client = Elasticsearch()
+
+    ubq = UpdateByQuery(using=client)
+
+You can also define the client at a later time (for more options see the
+:ref:`configuration` chapter):
+
+.. code:: python
+
+    ubq = ubq.using(client)
+
+.. note::
+
+    All methods return a *copy* of the object, making it safe to pass to
+    outside code.
+
+The API is chainable, allowing you to combine multiple method calls in one
+statement:
+
+.. code:: python
+
+    ubq = UpdateByQuery().using(client).query("match", title="python")
+
+To send the request to Elasticsearch:
+
+.. code:: python
+
+    response = ubq.execute()
+
+It should be noted, that there are limits to the chaining using the script method: calling script multiple times will 
+overwrite the previous value. That is, only a single script can be sent with a call. An attempt to use two scripts will 
+result in only the second script being stored.
+
+Given the below example:
+
+.. code:: python
+	
+	ubq = UpdateByQuery().using(client).script(source="ctx._source.likes++").script(source="ctx._source.likes+=2")
+
+This means that the stored script by this client will be ``'source': 'ctx._source.likes+=2'`` and the previous call
+will not be stored.
+
+For debugging purposes you can serialize the ``Update By Query`` object to a ``dict``
+explicitly:
+
+.. code:: python
+
+    print(ubq.to_dict())
+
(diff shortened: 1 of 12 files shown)


docs/update_by_query.rst
```diff
@@ -0,0 +1,148 @@
...
+For more information on queries, see the :ref:`search_dsl` chapter.
+
+Like the ``Search`` object, the API is designed to be chainable. This means that the ``Update By Query`` object
+is immutable: all changes to the object will result in a shallow copy being created which
+contains the changes. This means you can safely pass the ``Update By Query`` object to
+foreign code without fear of it modifying your objects as long as it sticks to
+the ``Update By Query`` object APIs.
+
+You can pass an instance of the low-level `elasticsearch client <https://elasticsearch-py.readthedocs.io/>`_ when
+instantiating the ``Update By Query`` object:
```
HonzaKral: can we instead point to the `configuration` chapter to promote global configuration instead of passing the client everywhere explicitly?

i know you mention it lower but that should be the recommended way and should be more highlighted.

elasticsearch_dsl/response/__init__.py
```diff
@@ -84,3 +84,16 @@ def __iter__(self):
...
+
+class UpdateByQueryResponse(AttrDict):
+
+    def __init__(self, search, response, doc_class=None):
+        super(AttrDict, self).__setattr__('_search', search)
+        super(AttrDict, self).__setattr__('_doc_class', doc_class)
+        super(UpdateByQueryResponse, self).__init__(response)
+
+    def __getitem__(self, attr_name):
+        return super(UpdateByQueryResponse, self).__getitem__(attr_name)
```
HonzaKral: I don't think this method is needed.

elasticsearch_dsl/update_by_query.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Return a clone of the current search request. Performs a shallow copy
+        of all the underlying objects. Used internally by most state modifying
+        APIs.
+        """
+        ubq = super(UpdateByQuery, self)._clone()
+
+        ubq._response_class = self._response_class
+        ubq._script = self._script.copy()
+        getattr(ubq, 'query')._proxied = getattr(self, 'query')._proxied
```
HonzaKral: `ubq.query._proxied = self.query._proxied` would be more readable here

elasticsearch_dsl/update_by_query.py
```diff
@@ -0,0 +1,149 @@
...
+
+        d.update(kwargs)
+        return d
+
+    def execute(self, ignore_cache=False):
+        """
+        Execute the search and return an instance of ``Response`` wrapping all
+        the data.
+        """
+        if ignore_cache or not hasattr(self, '_response'):
```
HonzaKral: I don't think we should ever attempt to cache the response to update by query as it is not a read-only operation like `Search`

test_elasticsearch_dsl/test_integration/test_update_by_query.py
```diff
@@ -0,0 +1,35 @@
...
+        .filter(~Q('exists', field='parent_shas'))\
+        .script(source='ctx._source.is_public = false')
+    ubq = ubq.params(conflicts='proceed')
+
+    response = ubq.execute()
+    assert response.total == 2
+    assert response.updated == 1
+    assert response.version_conflicts == 1
+
+def test_delete_by_query_with_script(data_client):
```
HonzaKral: This looks like a dangerous test - `data_client` uses the `git` and `flat-git` indices in a read-only fashion and it is therefore not torn down after every single test. Altering any data in those indices will have consequences in other tests and is to be avoided. Please use `write_client` and an index starting with `test-` for any test that alters documents in elasticsearch.

test_elasticsearch_dsl/test_update_by_query.py
```diff
@@ -0,0 +1,286 @@
+from copy import deepcopy
+
+from elasticsearch_dsl import UpdateByQuery, query, Q, Document
+
+
+# A lot of the basic query testing for Search can be copied over here and re-used
```
HonzaKral: there is no point in copying unit tests when your class uses the parent methods, it just introduces duplications. If you are not reimplmenting methods like `index()` do not copy their tests here.

emarcey: Thanks for your comments - I'm going to start working on them today.

But one question here: the UpdateByQuery doesn't inherit from Search because the query capabilities are so different - I can slim down the tests used here. Or would you rather I refactor to use UpdateByQuery inherits from Search to reduce duplication?

HonzaKral: True, but some of the tests here (like `.index()`) test the functionality of `Request` that `UpdateByQuery` does extend. Also it might be ok to create another class below `Request` that both `Search` and `UpdateByQuery` can then extend that would contain the common functionality, what do you think?

HonzaKral: Looking at it further it's just `query`, `filter`, and `exclude`. I would be ok with just a simple test each to verify they work since it's the same code as in `Search` where it is tested more extensively. Alternatively another class (either to extend or to mix in) is always a solution too.

emarcey:
@HonzaKral 

I think I addressed all of your comments. Let me know if there's something I missed.

Re: unit tests. For now, I slimmed down the tests to cover only what I have changed. I'm going to make a note to try and re-factor some time in the future

HonzaKral:
Thank you so much for the feature, @emarcey ! Much appreciated!

Sphinxs:
Does it support Python `dict` ? When I submit `script(source=f"ctx._source.test={dumps({'test': 1})}")` I receive: `TransportError(500, 'script_exception', 'compile error')`, but it works for strings, lists...

PR #1054, opened 2018-11-09
Fix location field typo when creating user

tcoyze:
Location field was misspelled in creating the users in the `parent_child.py` example.


HonzaKral:
Thank you very much for the fix, good catch!

Issue #725, opened 2017-08-27
Can't achieve autocomplete


kwasielew:
Hello, 
I am struggling with autocomplete for a long time.
Simply i am loading list of cities to the elasticsearch, i am printing first index so i am sure that Novinki city exists there. And when I do the execute suggest I receive  no completion for text "Novin".
I was refering to official tutorial - http://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html
Can someone give me a tip what I am doing wrong?
Thank you
```
import json
from elasticsearch_dsl import Search, Index, DocType, Text, Completion
from elasticsearch_dsl.connections import connections


connections.create_connection(hosts=['localhost'])

with open('city_list.json') as json_data:
    cities_json = json.load(json_data)

cities = Index('cities')
cities.delete(ignore=404)
cities.create()


@cities.doc_type
class City(DocType):
    name = Text()
    name_suggest = Completion()
    country = Text()

    class Meta:
        doc_type = 'city'

City.init()

for index, city in enumerate(cities_json['cities']):
    if index == 1:
        print(city) # this prints {'coord': {'lon': 37.666668, 'lat': 55.683334}, 'name': 'Novinki', 'id': 519188, 'country': 'RU'}
    city = City(name=city['name'], country=city['country'])
    city.save()

s = City.search()
s = s.suggest('name_suggestions', 'Novin', completion={'field': 'name_suggest'})
suggestions = s.execute_suggest()
for result in suggestions.name_suggestions:
    print('Suggestions for %s:' % result) # this gives me Suggestions for {'length': 5, 'options': [], 'offset': 0, 'text': 'Novin'}:

```

kwasielew:
OK, so I figured out that i was missing assigment to the name_suggest field:
`    city = City(name=city['name'], name_suggest=city['name'], country=city['country'])
`
The documentation lacks that information & there is no payload in suggestions anymore(still in documentation)

Still I have further question, how can I apply other analyzer to the Completion field? It is not able to do that like it can be in Text field.

HonzaKral:
Hi, yes, you can pass an `analyzer` property to the completion field, see https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-suggesters-completion.html#completion-suggester-mapping for more details.

Hope this helps!

princessRapunzel:
@HonzaKral how to set the analyzer property for Completion field through elasticsearch-dsl API?  
[...]

PR #1063, opened 2018-11-20
Fix Completion field

marcelometal:
Added analyzer param (Ref #725 )


HonzaKral:
Sorry about missing your original post on the issue @marcelometal , the patch is great, thank you for submitting it!

PR #1066, opened 2018-11-22
Add UpdateByQuery to the example

janitha09:
Add the missing UpdateByQuery to the import for the update example


HonzaKral:

PR #1078, opened 2018-11-29
Accept tuples as Date fields

seanthegeek:


HonzaKral:
What is the use case for this patch? `deserialize` is called when deserializing the data that came from elasticsearch. Since elasticsearch uses `json` which doesn't support `tuple`s this is not a valid data format for the value.

In `serialize` (assigning tuples from python) this is already supported.

seanthegeek:
Deserialize also takes place during a `save()`

Without this patch, this traceback happens:

```python
  File "C:\Users\Sean\GitHub\parsedmarc\parsedmarc\elastic.py", line 358, in save_aggregate_report_to_elasticsearch
    agg_doc.save()
  File "C:\Users\Sean\GitHub\parsedmarc\parsedmarc\elastic.py", line 94, in save
    return super().save(** kwargs)
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\document.py", line 425, in save
    self.full_clean()
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\utils.py", line 491, in full_clean
    self.clean_fields()
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\utils.py", line 477, in clean_fields
    data = field.clean(data)
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\field.py", line 99, in clean
    data = self.deserialize(data)
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\field.py", line 95, in deserialize
    return self._deserialize(data)
  File "C:\Users\Sean\GitHub\parsedmarc\venv\lib\site-packages\elasticsearch_dsl\field.py", line 259, in _deserialize
    raise ValidationException('Could not parse date from the value (%r)' % data)
TypeError: not all arguments converted during string formatting
```

With the patch, the dates get saved as a JSON list of ISO time strings, which Kibana treats as a date field with a range :) 

Also, I just noticed copy pasting threw off my Python indentation. I'll fix that right now.

HonzaKral:
The proper solution for this would be  #1079

but I am not sure we should support this functionality as there is no way to make it work cleanly as described for lists and also since elasticsearch (or json) doesn't support tuples. Will have to think if this can be implemented.

Thank you for the feedback, much appreciated. For now I would definitely recommend that you use lists instead as those always worked.

seanthegeek:
Using a list instead worked perfectly. Thanks!

PR #1095, opened 2018-12-25
Update aggs.py weighted average aggregation

truongvan:
add class support weighted average aggregation. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-weight-avg-aggregation.html


HonzaKral:
Thank you very much for the patch, @truongvan!

PR #1098, opened 2018-12-30
Fix suggest query names for Completion suggestor.

tell-k:
According to [official documentation,](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html) Completion Suggester seems to be using correct `prefix` rather than `text`.

```
POST music/_search?pretty
{
    "suggest": {
        "song-suggest" : {
            "prefix" : "nir", 
            "completion" : { 
                "field" : "suggest" 
            }
        }
    }
}
```

In the `Search.suggest` method, however, it always becomes `text`, so I fixed it.

In Completion Suggestor you can also specify `regex`.

```
POST music/_search?pretty
{
    "suggest": {
        "song-suggest" : {
            "regex" : "n[ever|i]r",
            "completion" : {
                "field" : "suggest"
            }
        }
    }
}
```
I also modified it so that the user can select it.

```python
# Sample of use

# Term Suggester
s = s.suggest('song-suggest', 'nir', term={'field': 'suggest'})

# Completion Suggester + prefix query
s = s.suggest('song-suggest', 'nir', completion={'field': 'suggest'})

# Completion Suggester + regex query
s = s.suggest('song-suggest', regex='n[ever|i]r', completion={'field': 'suggest'})
# "regex"  must be keyword argument.
```

Please feel free to merge it. thx.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -605,7 +605,7 @@ class Search(Request):
             s._highlight[f] = kwargs
         return s
 
-    def suggest(self, name, text, **kwargs):
+    def suggest(self, name, text=None, regex=None, **kwargs):
         """
         Add a suggestions request to the search.
 
@@ -616,9 +616,23 @@ class Search(Request):
 
             s = Search()
             s = s.suggest('suggestion-1', 'Elasticsearch', term={'field': 'body'})
+
         """
+        if text is None and regex is None:
+            raise ValueError('You have to pass "text" or "regex" argument.')
+
         s = self._clone()
-        s._suggest[name] = {'text': text}
+
+        query_name = 'text'
+        query_text = text
+        if 'completion' in kwargs:
+            if text:
+                query_name= 'prefix'
+            elif regex:
+                query_name= 'regex'
+                query_text = regex
+
+        s._suggest[name] = {query_name: query_text}
         s._suggest[name].update(kwargs)
         return s
 
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -611,29 +611,33 @@ class Search(Request):
 
         :arg name: name of the suggestion
         :arg text: text to suggest on
+        :arg regex: regex query for Completion Suggester
 
         All keyword arguments will be added to the suggestions body. For example::
 
             s = Search()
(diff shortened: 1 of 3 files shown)


tell-k:
@HonzaKral

Is there any problem with this PR?
If there are problems, I'll fix them.

thx :)

elasticsearch_dsl/search.py
```diff
@@ -616,9 +616,23 @@ def suggest(self, name, text, **kwargs):
 
             s = Search()
             s = s.suggest('suggestion-1', 'Elasticsearch', term={'field': 'body'})
+
         """
+        if text is None and regex is None:
+            raise ValueError('You have to pass "text" or "regex" argument.')
```
HonzaKral: I think we need additional validation rules here - if you supply both `text` and `regex` or you supply `regex` and `completion` is not in `kwargs` etc - any invalida combination essentially should raise a `ValueError`.

elasticsearch_dsl/search.py
```diff
@@ -616,9 +616,23 @@ def suggest(self, name, text, **kwargs):
...
+        query_name = 'text'
+        query_text = text
+        if 'completion' in kwargs:
+            if text:
+                query_name= 'prefix'
+            elif regex:
+                query_name= 'regex'
+                query_text = regex
+
+        s._suggest[name] = {query_name: query_text}
```
HonzaKral: once we have validated out all the invalid combinations we should be ok with simpler code:
```python
d = s._suggest[name] = kwargs
if text:
    d['text'] = text
else:
    d['regex'] = regex
```

elasticsearch_dsl/search.py
tell-k: OK, i'll fix it more strictly.

tell-k:
@HonzaKral  

Thank you for your review.
I fixed  the points pointed out by you.
Please feel free to review it again.

I pray that you can afford the time :)

tell-k:
@HonzaKral 

Can you review this PR again ?
If there's anything I can do for you, please let me know.  Thx :)

tell-k:
@HonzaKral Any chance you can review and merge this?

sethmlarson:
Closing this PR as it's gone stale, rebase on master and I'll review again

Issue #1101, opened 2019-01-07
Support for Decimal field


AndreiHondrari:
There is an active issue in a repository that wraps around elasticsearch-dsl for use with django.
https://github.com/sabricot/django-elasticsearch-dsl/issues/57

It requires a serialization/deserialization using Decimal.

HonzaKral:
Thank you for the issue. In the mean time you can always implement the `DecimalField` on your own by using the `CustomField` helper as showcased in tests - https://github.com/elastic/elasticsearch-dsl-py/blob/master/test_elasticsearch_dsl/test_document.py#L48-L57

HonzaKral:
Coming back to this I am curious what would be the use case for this - elasticsearch itself doesn't have support for arbitrary precision floating point numbers beyon `float` and `double`. In that case would you still like to store `Decimal` numbers in elasticsearch? What kind of operations would you be looking for that don't work right now?

When I try:
```python
from elasticsearch_dsl import Document, Double
from decimal import Decimal
class D(Document):
    d = Double()
D(d=Decimal('0')).to_dict()
```

everything works as expected.

HonzaKral:
I am closing this as there is no additional info on expectations. Also Elasticsearch itself doesn't support arbitrary precision numbers, only `float`, `double`, and `scaled_float`.

Please feel free to reopen with additional information

PR #1102, opened 2019-01-07
[Implement #1101] Add support for Decimal field.

AndreiHondrari:


HonzaKral:
I am sorry but I cannot accept this PR. Elasticsearch itself doesn't have a `Decimal` field, only more standard numeric fields (0) are supported. And since serialization of `Decimal` is already supported by the underlying `elasticsearch-py` library I don't see the need for this functionality.

Thank you


0 - https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html

PR #1104, opened 2019-01-07
[Fix #1103] Replace dummy_response fixture call with a constant.

### LINKED ISSUE #1103: PyTest 4.x.x no longer supports calling fixtures directly
In the latest master if you execute `python setup.py test` it will yield a bunch of fixture errors 

> __ ERROR at setup of test_save_no_index __
> __ ERROR at setup of test_delete_no_index __
> __ ERROR at setup of test_search_with_custom_alias_and_index ___
> __ ERROR at setup of test_cache_can_be_ignored __
> __ ERROR at setup of test_params_being_passed_to_search __
> __ ERROR at setup of test_delete_by_query __
> __ ERROR at setup of test_params_being_passed_to_search __

all with the same message: 

> Fixture "dummy_response" called directly. Fixtures are not meant to be called directly,
> but are created automatically when test functions request them as parameters.
> See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
> https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.

All being triggered by https://github.com/elastic/elasticsearch-dsl-py/blob/7ece785db6367da7ace00b32aa0258d4bf4baf28/test_elasticsearch_dsl/conftest.py#L35

It is recommended to replace this with a constant like `DUMMY_RESPONSE_DATA` which should be returned by the `dummy_response` fixture and used in the `mock` fixture.

AndreiHondrari:


HonzaKral:
thank you for the patch, I decided to go with a different direction in b104550 but this issue was very helpful in immediately seeing what's wrong.

PR #1116, opened 2019-02-06
Fix tiny typo in documentation

sankalp-khare:
It says "reach" where it should say "each". Thanks!


sankalp-khare:
Filing a fresh one after signing the contributor agreement.

PR #1117, opened 2019-02-06
Fix tiny typo in documentation

sankalp-khare:
Says "reach" where it should say "each".


sankalp-khare:
I did sign the CLA before filing this pull request. Am I missing something?

HonzaKral:
Thank you for the PR, much appreciated. We were having some issues with the CLA checker so that's why the false negative check...

PR #1131, opened 2019-03-01
[Fix #1084] Fix Hit object to pickle the meta member

### LINKED ISSUE #1084: Hit objects are not pickleable
`Hit` obejcts are not properly serialized when using `pickle` because the `meta` member is not available after loading the object using `pickle.load`. This happens because of the [way ](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/response/hit.py#L13) the `meta` member of the `Hit` class is added in the `__init__` method. Anyways, I have fixed it by adding `__getstate__` and `__setstate__` methods to `Hit` that make sure we don't lose the metadata. I have the fix on a local fork and would like to open a PR :)

eli314:
This fixes [#1084](https://github.com/elastic/elasticsearch-dsl-py/issues/1084).

I would also like to to the same for the 5.x branch as this is the branch we use in our company, I'll open a new PR once this one is merged.

Thanks :)

### DIFF
--- a/elasticsearch_dsl/response/hit.py
+++ b/elasticsearch_dsl/response/hit.py
@@ -12,6 +12,14 @@ class Hit(AttrDict):
         # assign meta as attribute and not as key in self._d_
         super(AttrDict, self).__setattr__('meta', HitMeta(document))
 
+    def __getstate__(self):
+        # add self.meta since it is not in self.__dict__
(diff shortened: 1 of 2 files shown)


HonzaKral:
Thank you so much for the PR, much appreciated!

PR #1132, opened 2019-03-01
[Fix #1084]Fix Hit object to pickle the meta member in 5.x versions

### LINKED ISSUE #1084: Hit objects are not pickleable

elilutsky:
As stated [here](https://github.com/elastic/elasticsearch-dsl-py/pull/1131) I'm opening a similar PR to fix the [issue ](https://github.com/elastic/elasticsearch-dsl-py/issues/1084) for 5.x since we use this version in our company.

Thanks :)

### DIFF
--- a/elasticsearch_dsl/response/hit.py
+++ b/elasticsearch_dsl/response/hit.py
@@ -22,6 +22,14 @@ class Hit(AttrDict):
         # assign meta as attribute and not as key in self._d_
         super(AttrDict, self).__setattr__('meta', HitMeta(document))
 
+    def __getstate__(self):
+        # add self.meta since it is not in self.__dict__
+        return super(Hit, self).__getstate__() + (self.meta, )
+
+    def __setstate__(self, state):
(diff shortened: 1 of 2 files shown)


eli314:
The CI currently fails on the original 5.x branch. This is not related to the PR.

elilutsky:
@HonzaKral I'd be happy if you could CR this so we can apply this patch in our company

sethmlarson:
The 5.x branch isn't supported any longer and likely won't see a release. Closing this PR.

PR #1148, opened 2019-03-25
Reintroduce Attachment Field

### LINKED ISSUE #1114: Missing Attachment api in the elasticsearrch_dsl.field module
I am unable to find the Attachement Api in the elasticsearch_dsl.field module, how do I pass the attahment in the IngestClient ?

murphydavis:
This change fixes #1114, which seemed to arise from the (unintentional?) removal of the Attachment field in a merge commit 21a620d758d13d1ca3d20c36d6db0fead0e8b5fe

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -359,6 +359,9 @@ class Binary(Field):
             return None
         return base64.b64encode(data).decode()
 
+class Attachment(Field):
+    name = 'attachment'
+
 class GeoPoint(Field):
     name = 'geo_point'
 


elasticcla:
Hi @murphydavis, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/1148.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

HonzaKral:
The removal wasn't accidental, elasticsearch `6.x` and onwards doesn't have an `attachment` data type. See https://www.elastic.co/guide/en/elasticsearch/reference/6.7/mapping-types.html

Thank you for the PR

PR #1150, opened 2019-03-28
Cleans up minor issues for pep/linter/doc

3lnc:


HonzaKral:
Thanks for the patch, @3lnc !

PR #1151, opened 2019-03-28
Fix FacetedSearch with ElasticSearch 6.4

debuggerpk:
If we don't provide the fields, ElasticSearch itself provides fields as `["*"]` and we do not provide it exclusively. Setting the fields exclusively to `('*',)`, and doing search without providing the index, `multi_search` breaks. 
This PR fixes that.


HonzaKral:
Thank you for the PR, much appreciated!

PR #1184, opened 2019-05-12
Don't override existing extras in update_from_dict

konstin:
Previously, update_from_dict would discard existing extras, e.g. 
```
search = Search()
search.update_from_dict({"indices_boost": [{"important-documents": 2}]})
search.update_from_dict({"_source": ["id", "name"]})
```
made `search.to_dict()` equal `{'_source': ['id', 'name']}`. With this change, `indices_boost` is kept in extras together with `_source`.


HonzaKral:

PR #1185, opened 2019-05-14
Fix typos in docstrings

dsanders11:


HonzaKral:
Thank you for the patch, @dsanders11 !

PR #1192, opened 2019-05-16
Drop deprecated DocType

dsanders11:
As mentioned in the docs and the comment in the code, `DocType` is meant to be dropped in 7.0.0, but still exists at the moment.


HonzaKral:
Good catch, thanks for the patch, @dsanders11 !

PR #1197, opened 2019-05-20
Optimistic concurrent control for Document.delete

dsanders11:
Optimistic concurrency is also available with the Delete API. Not sure if it was skipped on purpose or simply an oversight.


HonzaKral:
Definitely an oversight, thank you for the pr, @dsanders11!

PR #1213, opened 2019-06-26
Refactores usage of connection singeton to exposed functions

3lnc:
Since refactoring interface of `connections`, we now have more clean way to use connections machinery.

PR provides small refactoring on this, plus import order fixes for affected fields.

AFAIK, this is fully backward-compatible change, can be released with next minor version.


3lnc:
ping @HonzaKral

HonzaKral:
Thank you again for the ping, I appreciate it!

PR #1214, opened 2019-06-26
Fixes count cache logic according to 7.x API

3lnc:
ES 7.x API breaks current implementation of count cache for `Search`.

Adds corresponding integration test. Removed unit-test for it — IMO this gives false coverage.


3lnc:
ping @HonzaKral

HonzaKral:
thank you, @3lnc !

PR #1217, opened 2019-07-02
Index.create and IndexTemplate.save returns raw results

3lnc:
- unifies .save and .create to follow overall "proxy methods" interface schema (this also gives ability to easily make async dsl)
- removes default doctype constant. This is no longer used, and never was public interface — @HonzaKral please correct me if I'm wrong on this
- pep fixes on affected files


3lnc:
ping @HonzaKral

HonzaKral:
thank you for the ping and, of course, the PR!

PR #1255, opened 2019-10-20
Prevent Object/Nested from mutating the inner doc_class' mapping

l1nd3r0th:
If the same inner doc is used with both Object and Nested they override each other's mapping type, as shown in the test.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -1,4 +1,5 @@
 import base64
+import copy
 import ipaddress
 
 try:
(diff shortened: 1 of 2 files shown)


twiddli:
This is a pretty subtle but serious issue. I just encountered it. Can this get merged?

l1nd3r0th:
> This is a pretty subtle but serious issue. I just encountered it. Can this get merged?

I'm all for!

Maintainers, if there's something I can help with to get this moving, please let me know.

PR #1262, opened 2019-10-23
Fix connection configuration example syntax

eric-weaver:
Updates the example code with correct syntax. This change now aligns with example documentation in the docs that was updated [here](https://github.com/elastic/elasticsearch-dsl-py/commit/a81d1288e2d51c6d9003d913f283646352fe0255)


HonzaKral:
Thank you for the patch, @eric-weaver, much appreciated!

PR #1263, opened 2019-10-24
Fix possessive pronoun typo

rcjsuen:
"it's" is a contraction of "it is" so we should use the possessive pronoun of "its" instead.

### DIFF
--- a/elasticsearch_dsl/document.py
+++ b/elasticsearch_dsl/document.py
@@ -170,7 +170,7 @@ class Document(ObjectBase):
     @classmethod
     def get(cls, id, using=None, index=None, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


HonzaKral:
Thank you for the fix, @rcjsuen, much appreciated!

rcjsuen:
@HonzaKral My changes were not merged and I don't see anything in `master`. Was there something wrong with my suggested change?

HonzaKral:
D'oh, not only is my English patchy at times, I also pushed the wrong button and thought I merged your PR when I actually just closed it, my bad!

Thank you very much for checking, it should be ok now!

rcjsuen:
Thanks for checking, @HonzaKral. No worries and thank you for reviewing the pull request. Have a nice day!

PR #1295, opened 2020-01-11
Issue #1292: Add search_as_you_type datatype

### LINKED ISSUE #1292: Implementation of search_as_you_type field mapping and query
I've noticed that the recently added [search_as_you_type](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-as-you-type.html) field is not implemented in this package. Would it be of interest of the community to add it? I'd be happy to try and create a PR for it.

Also, elasticsearch itself doesn't provide a specific search query for this field and suggests the use of multi_match, as follows:

```
{
  "query": {
    "multi_match": {
      "query": "brown f",
      "type": "bool_prefix",
      "fields": [
        "my_field",
        "my_field._2gram",
        "my_field._3gram"
      ]
    }
  }
}
```
Would it be nice to have a SearchAsYouType query class in elasticsearch-dsl that generates this multi_match query?

dpasqualin:
Close #1292 by adding a class and example on how to use the search_as_you_type datatype.
Example is very similar to completion.py, I hope it is ok.

### DIFF
--- a/elasticsearch_dsl/field.py
+++ b/elasticsearch_dsl/field.py
@@ -266,6 +266,14 @@ class Text(Field):
     }
     name = 'text'
 
+class SearchAsYouType(Field):
+    _param_defs = {
+        'analyzer': {'type': 'analyzer'},
+        'search_analyzer': {'type': 'analyzer'},
+        'search_quote_analyzer': {'type': 'analyzer'},
+    }
+    name = 'search_as_you_type'
+
 class Keyword(Field):
     _param_defs = {
         'fields': {'type': 'field', 'hash': True},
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @dpasqualin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-dsl-py/pull/1295.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

zhanwenchen:
@HonzaKral Would you please take a look? I applied this PR to my fork but I'm really hoping to use this feature. It is now standard in 7.2.

HonzaKral:
Thank you for the ping, @zhanwenchen !

PR #1305, opened 2020-01-30
FacetedSearch missing 'using' parameter

isoboroff:
The FacetedSearch class has an internal member 'using' which is needed for the Search call, but there's no way to pass a client in for it to use, so it only works with 'default'.


isoboroff:
Oh god I am not signing a contributer agreement for a one-liner fix.  I hereby license my fix under the original license of the code.  ffs.

sethmlarson:
Closing since we require a signed CLA for all changes, sorry about that.

PR #1317, opened 2020-02-21
Modified Document delete method to return result

jesserobles:
The Document.delete method currently returns None, which is not consistent with the save and update methods. It also makes checking delete success and running test a little awkward. Also updated docstring in save method to be accurate.


sethmlarson:
Hey! Thanks for this change. It seems like this PR is a bit stale. I'll be closing this until the CLA is signed, test cases are added, and the branch is brought up to date with master branch.

PR #1346, opened 2020-04-22
Add python 3.8 support

jaywhy13:
This PR adds support for Python 3.7 and 3.8


jaywhy13:
> LGTM, thank you!

My pleasure. How does your release cycle work? This is a change I was really hoping to integrate.

brainix:
@jaywhy13 - It looks like the only real functional change is pinning the version of `pytest-mock`? Could you temporarily do that in your Python project's `requirements.txt` file until the next release of `elasticsearch-dsl` is cut?

jaywhy13:
> Edit: For what it's worth, I've been using `elasticsearch-dsl` on Python 3.8 in production for months, even without pinning `pytest-mock`.

Yeah, I'm actually using Poetry and it's too opinionated to let me install a package that doesn't have support for Python 3.8. I'll likely have to roll my own until this is available. I'm upgrading to ES 7, but also did a Python 3.8 bump as well.

PR #1353, opened 2020-04-29
add __len__ to support django paginator

teddyinfi:


sethmlarson:
Could this functionality be added to Django-Elasticsearch-DSL directly if it is desired there? Otherwise users are free to call this themselves.

Issue #1358, opened 2020-05-05
Reverse slices result in negative size


bk-equityzen:
```py
search[20:0]
```
results in `-20` as the size sent to ElasticSearch. This can possibly cause an error: `RequestError(400, 'search_phase_execution_exception', 'numHits must be > 0; please use TotalHitCountCollector if you just need the total hit count')`

Proposed Solution:
The slice function should result in a minimum size of `0`

```py
s._extra['size'] = n.stop - (n.start or 0) if n.stop is not None else 10
s._extra['size'] = max(s._extra['size'], 0)
```

sethmlarson:
Nice catch! Would you be willing to provide a fix and test case for this issue?

bk-equityzen:
> Nice catch! Would you be willing to provide a fix and test case for this issue?

Done! https://github.com/elastic/elasticsearch-dsl-py/pull/1360

shanon-equityzen:
👋  @sethmlarson thanks for merging this in -- can you please let us know when we can expect this to be released?  🙌

sethmlarson:
@shanon-equityzen I can try to get a patch release out tomorrow.

bk-equityzen:
@sethmlarson any update on the next release?

sethmlarson:
@bk-equityzen Released v7.2.1

PR #1360, opened 2020-05-05
Avoid negative sizes in Search slice

bk-equityzen:
Resolves https://github.com/elastic/elasticsearch-dsl-py/issues/1358


elasticsearch_dsl/search.py
```diff
@@ -345,6 +345,7 @@ def __getitem__(self, n):
             # stop not given.
             s._extra['from'] = n.start or 0
             s._extra['size'] = n.stop - (n.start or 0) if n.stop is not None else 10
+            s._extra['size'] = max(s._extra['size'], 0)
```
sethmlarson, line 348: Can we move the `max(X, 0)` call on the same line as where `_extra['size']` is calculated above?

PR #1370, opened 2020-06-09
Remove extra nesting level in generated query for Search.exclude

dan-blanchard:
As you can see from the updated test, `Search.exclude` previously generated a query that had a `bool` nested inside a `bool`, instead of using `must_not`. This makes the generated queries much easier to read, especially when you've made multiple calls to `Search.filter` first.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -314,7 +314,7 @@ class Search(Request):
         return self.query(Bool(filter=[Q(*args, **kwargs)]))
 
     def exclude(self, *args, **kwargs):
-        return self.query(Bool(filter=[~Q(*args, **kwargs)]))
+        return self.query(~Q(*args, **kwargs))
 
     def __iter__(self):
         """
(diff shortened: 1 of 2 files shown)


dan-blanchard:
I signed the CLA now

sethmlarson:
Going to close this PR as stale. I'm also not sure if there are any benefits to changing this now.

dan-blanchard:
What's stale about it?

sethmlarson:
The CLA bot didn't update the status since the PR was created leading me to think that this PR had been abandoned at that step, my mistake. I still am not sure about the change itself, I think I'd rather leave things as they are if they are identical functionally.

dan-blanchard:
The functionality isn't really identical. ES may treat them the same, but one is substantially more human readable. Especially if you make multiple calls to `exclude`. Without this change each call to `exclude` makes another nested `bool`. Beyond being hard to read, if the query is complex and frequent enough, this can actually noticeably impact bandwidth usage.

PR #1375, opened 2020-06-16
docs: fixed example with hits

alpolishchuk:
Fixed example "Hits" in docs


alpolishchuk:
@sethmlarson CLA signed, can we merge this?

PR #1288, opened 2019-12-31
Remove standard filter

truongvan:
Standard token filter has been removed in 7.0.

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#standard-filter-removed

PR #1436, opened 2020-10-05
Do not use "standard" filter when testing against 7.x

lolripgg:
The "standard" filter has been removed in 7.x and causes the integration tests to fail when run against Elasticsearch 7.x.

Per @HonzaKral's comment on #1288, 7.x clusters can still have indices created in 6.x, when the "standard" filter was valid. To continue supporting this until an 8.x release, this commit checks the Elasticsearch version the tests are being run against before constructing the analyzer that currently uses the "standard" filter. If the Elasticsearch version is >= 7, the "standard" filter is not used.

### DIFF
--- a/test_elasticsearch_dsl/test_integration/test_document.py
+++ b/test_elasticsearch_dsl/test_integration/test_document.py
@@ -15,10 +15,12 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
+import os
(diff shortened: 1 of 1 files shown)


sethmlarson:
Issue was resolved in https://github.com/elastic/elasticsearch-dsl-py/pull/1457

PR #1444, opened 2020-10-28
Check for failed shards in count

### LINKED ISSUE #1302: Count returns 0 and does not check for failed shards
When performing count with an illegal query (too many clauses, invalid regex, etc'), the error is hidden from the user and the return value is 0, which is wrong.

For example if we run an invalid query, with date range filtering on several indices, the result will include "successful" shards which are considered successful because they did not meet the date range criteria, and some failed shards for indices who did meet the range, but failed because of the query.

The response is not validated and there is no check for failed shards. Actually there is a 6-years old TODO that mentions that:

image

Here is an example response:
```
{'count': 0,
 '_shards': {'total': 12,
  'successful': 6,
  'skipped': 0,
  'failed': 6,
  'failures': [{'shard': 0,
    'index': 'my-index-2020-01-20',
    'node': 'fHCBfQdfaraIDaqOH-i2Ug',
    'reason': {'type': 'query_shard_exception',
```

and the return value will be 0.

This behaviour is confusing and unwanted, as you can search for important events and get '0' from the count, although there was just an error in the query.

yogevyuval:
Fixes #1302 .

When some shards fail, the count is returned as a default or a partial answer, hiding the error from the user.


yogevyuval:
@sethmlarson

yogevyuval:
@tommyzli

yogevyuval:
@armando1793

yogevyuval:
@hugovk

yogevyuval:
@pquentin

miguelgrinberg:
Unfortunately it is difficult to know how to handle shard failures. Automatically raising a failure is not always the correct option, to me it feels the best option would be to expose the information, so that each application can decide if and how many shards failing should be considered a failure.

Based on your description in #1302 you are sending queries that are bad, so in that case a returned count of 0 does not seem completely wrong.

So I'm going to close this PR, but will keep the issue open, as we need a way to expose the complete results of the count operation for applications that need to make decisions based on that.

PR #1476, opened 2020-12-18
Add a comment about accessing the low level connection client

Whatapalaver:
Edited the configuration documentation to add a comment about accessing the low level connection client which can be useful when working with some elasticsearch_py methods


sethmlarson:
Thanks for opening this but I think we're going to pass on this. Users that require additional methods should use their Elasticsearch client directly instead of going through Elasticsearch-DSL.

Issue #1501, opened 2021-05-02
Add 'year' to date intervals for DateHistogramFacet


nrsimha:
There is missing `year` interval in DateHistogramFacet.

nrsimha:
Added pull request #1502

PR #1502, opened 2021-05-02
Add year interval for DateHistogramFacet

nrsimha:
Added `year` interval for DateHistogramFacet.

Closes https://github.com/elastic/elasticsearch-dsl-py/issues/1501

### DIFF
--- a/elasticsearch_dsl/faceted_search.py
+++ b/elasticsearch_dsl/faceted_search.py
@@ -168,6 +168,10 @@ class HistogramFacet(Facet):
         )
 
 
+def _date_interval_year(d):
+    return (d + timedelta(days=366)).replace(day=1)
+
+
 def _date_interval_month(d):
     return (d + timedelta(days=32)).replace(day=1)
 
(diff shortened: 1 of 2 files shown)


nrsimha:
Is there anything I can do to make this merged?

elasticsearch_dsl/faceted_search.py
```diff
@@ -168,6 +168,10 @@ def get_value_filter(self, filter_value):
         )
 
 
+def _date_interval_year(d):
+    return (d + timedelta(days=366)).replace(day=1)
```
sethmlarson, line 172: Let's change this to `return d.replace(year=d.year+1, day=(28 if d.month == 2 and d.day == 29 else d.day))` since years are linear unlike months.

We should add a test case for leap day handling as well (Basically see that February 29th, 2020 increments to February 28th, 2021)

sethmlarson (changes requested):
Thanks for this, one comment for you:

elasticsearch_dsl/faceted_search.py
nrsimha: Done. Updated to ```return d.replace(year=d.year+1, day=(28 if d.month == 2 and d.day == 29 else d.day))```

sethmlarson (approved):
LGTM

nrsimha:
Thank you

PR #1509, opened 2021-06-09
Pin elasticsearch-py dependency to a specific version

### LINKED ISSUE #412: Distribute elasticsearch-dsl-py 1.x and elasticsearch-dsl-py 2.0 separately
Currently elasticsearch-dsl is distributed on python as one package with different versions, which means that it is impossible to have both elasticsearch 1.x and 2.0 installed at the same time, although it might make very much sense for the same application to talk to different versions of elasticsearch.

From what I have gathered in my research it is more or less impossible to get multiple versions of the same project using pip or some similar version. So what I want to propose is the following:

Instead of having one package with two versions two packages should be maintained on the package index, elasticsearch-dsl and elasticsearch-dsl2, which would enable people to use both versions at the same time.

I'd be very interested in hearing your opinions on the matter!

### LINKED ISSUE #919: Consider release multiple packages for different major versions?
Underlying elasticsearch-py is doing this, there are elasticsearch2 elasticsearch5 packages on pypi. Should elasticsearch-dsl do the samething? We have a relative large code base and multiple versions of elasticsearch clusters, this will make our lives much easier.

potasiak:
To make it possible to install the package for multiple Elasticsearch
versions in a single Python environment, the name of the provided
Python package has been changed from `elasticsearch_dsl` to
`elasticsearch2_dsl`, along with its PyPI name from `elasticsearch-dsl`
to `elasticsearch2-dsl`.

The dependency on `elasticsearch` Python package has been changed to
dependency on `elasticsearch2` package.

Closes #412
Closes #919


potasiak:
There are issues with Travis CI - some Python versions could not be downloaded (2.6, 3.3), 3.4 raises an exception:
```
    NOTSET: "Final" = NotSetType.token  # noqa: E305
          ^
SyntaxError: invalid syntax
```

Should I update the CI for the `2.x` branch to GitHub workflows first?
Should I change supported Python versions to exclude ones that are long gone now (2.6 and 3.3)?

sethmlarson:
The 2.x version of this project is no longer receiving updates so I don't think we can accept this PR.

PR #1513, opened 2021-07-05
Changed the folder name

FultonSlide:
Changed the folder name.


sethmlarson:
We don't want to change the directory name, we build the package so it doesn't need this type of aliasing. Also, Elasticsearch v5 isn't supported any longer.

FultonSlide:
> We don't want to change the directory name, we build the package so it doesn't need this type of aliasing. Also, Elasticsearch v5 isn't supported any longer.

Apologies didn't mean to make this, this was an accidental PR. Thanks for closing.

PR #1536, opened 2021-09-29
Fix merge source data of `utils.merge` is `AttrDict` but not an iterable

### LINKED ISSUE #1535: `Document.update` doesn't allow updates with `InnerDoc`
For example:

```python
class Location(InnerDoc):
  country = Text()
  city = Text()
  
class User(Document):
  name = Text()
  location = Object(Location)

  class Index:
    ...

user = User(name='alice', location=Location(country='foo', city='bar'))
user.save()
user.update(location=Location(city='baz'))
```

then it will raise the following error

image

`InnerDoc` inherits from `AttrDict`, and [`merge`](https://github.com/elastic/elasticsearch-dsl-py/blob/v7.4.0/elasticsearch_dsl/utils.py#L547) method accepts `AttrDict` to be merged, the problem is that not all `AttrDict` subclasses implement a dict-like interface for [iterate](https://github.com/elastic/elasticsearch-dsl-py/blob/v7.4.0/elasticsearch_dsl/utils.py#L558), such as `InnerDoc`

I think, its better to first convert `AttrDict` to dict in `merge` method to make it iterable. After this fix, doc creation & update can both accept `InnerDoc`.

kaiix:
resolves #1535

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -555,6 +555,9 @@ def merge(data, new_data, raise_on_conflict=False):
             )
         )
 
+    if isinstance(new_data, AttrDict):
+        new_data = new_data.to_dict()
+
     for key, value in iteritems(new_data):
         if (
             key in data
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Given that the `AttrDict` class already implements most of the dict interface I think it makes more sense to add `items()` there. Adding a fix for that in #1784

PR #1584, opened 2022-04-27
Add AttrList.to_list()

kujiy:
`AttrDict` has `to_dict()` but `AttrList` does not have `to_list()`.

`AttrList` does not work for PyDantic as a data type. It's good if anyone can convert `AttrList` to `list` easily.

Thanks,

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -102,7 +102,8 @@ class AttrList:
         self._l_, self._obj_wrapper = state
 
     def to_list(self):
-        return [x for x in self.__iter__()]
+        return self._l_
+
 
 class AttrDict:
     """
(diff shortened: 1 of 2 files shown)


kujiy:
I signed the CLA. Please proceed the review, thanks

elasticsearch_dsl/utils.py
```diff
@@ -101,6 +101,8 @@ def __getstate__(self):
     def __setstate__(self, state):
         self._l_, self._obj_wrapper = state
 
+    def to_list(self):
+        return [x for x in self.__iter__()]
```
eapermyakov, line 105: I suggest doing it analogically with AttrDict. 

This will get the original value.
```
def to_list(self):
    return self._l_
```

miguelgrinberg (approved):

PR #1597, opened 2022-05-23
Switch from mock to unittest.mock

### LINKED ISSUE #1596: use unittest.mock instead of mock
Could you please consider to use unittest.mock instead of mock, as a fallback at least?

https://github.com/elastic/elasticsearch-py/issues/1983
https://trello.com/c/S6eADbii/64-remove-python-mock
https://fedoraproject.org/wiki/Changes/DeprecatePythonMock

It works for 3.6 and up at least:
https://build.opensuse.org/package/show/home:pgajdos:python/python-elasticsearch-dsl

pquentin:
Closes #1596

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,6 @@ install_requires = [
 ]
 
 develop_requires = [
-    "mock",
     "pytest>=3.0.0",
     "pytest-cov",
     "pytest-mock<3.0.0",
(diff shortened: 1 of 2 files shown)


pquentin:
Right, so if we do intend to support 2.7, then this is going to break support.

PR #1653, opened 2023-07-09
Docs: Clarify paragraph about persistent data types

tienne-B:
A paragraph about data types used in creating persistant data models is really confusing as it talks about instantiating while the examples given do not instantiate the classes. This commit rewords the paragraph for clarity and precision on how to use `Document` and `InnerDoc` and their effect.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -66,10 +66,11 @@ settings in elasticsearch (see :ref:`life-cycle` for details).
 Data types
 ~~~~~~~~~~
 
-The ``Document`` instances should be using native python types like
-``datetime``. In case of ``Object`` or ``Nested`` fields an instance of the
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
Thanks. I'm reviewing your suggested changes, and while I agree that the current documentation could provide more details I think your edits do not add clarity and instead introduce additional points of potential confusion, such as discussing how object and nested fields are declared, even though the matter discussed in this paragraph is how `Document` fields are used. You also suggest above that the example referenced in this paragraph does not instantiate any classes and this is incorrect, the example creates a `Comment` instance.

I think this section of the documentation needs to be expanded, including a separate section or paragraph on how to declare fields (both simple types and `InnerDoc` instances), possibly referencing specific sections of the included example.

PR #1660, opened 2023-08-21
Updated metadata to use group email address

technige:
This PR removes individual names and email addresses and replaces those with the new (external) [group email address](mailto:client-libs@elastic.co).

The author and maintainer fields have also been collapsed down into just author. According to [setuptools](https://setuptools.pypa.io/en/latest/references/keywords.html), the maintainer is used as an override for author, so the former is removed to use only the primary field.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,10 +50,8 @@ setup(
     long_description=long_description,
     long_description_content_type="text/x-rst",
     version=__versionstr__,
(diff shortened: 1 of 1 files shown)


JoshMock (approved):

JoshMock:
3.6 tests failed for unrelated reasons.

PR #1664, opened 2023-08-28
Support Elasticsearch 8

pquentin:
When I say "support", I mean that all tests are passing with >90% coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
           "3.10",
           "3.11",
         ]
-        es-version: [7.0.0, 7.10.0]
+        es-version: [8.0.0, 8.9.0]
 
     steps:
       - name: Checkout Repository
@@ -76,7 +76,7 @@ jobs:
         run: |
           mkdir /tmp/elasticsearch
           wget -O - https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${{ matrix.es-version }}-linux-x86_64.tar.gz | tar xz --directory=/tmp/elasticsearch --strip-components=1
-          /tmp/elasticsearch/bin/elasticsearch -d
+          /tmp/elasticsearch/bin/elasticsearch -E xpack.security.enabled=false -E discovery.type=single-node -d
       - name: Setup Python - ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -18,7 +18,7 @@
 import collections.abc
 import copy
 
-from elasticsearch.exceptions import TransportError
+from elasticsearch.exceptions import ApiError
 from elasticsearch.helpers import scan
 
 from .aggs import A, AggBase
@@ -693,7 +693,8 @@ class Search(Request):
 
         d = self.to_dict(count=True)
         # TODO: failed shards detection
-        return es.count(index=self._index, body=d, **self._params)["count"]
+        resp = es.count(index=self._index, query=d.get("query", None), **self._params)
+        return resp["count"]
 
     def execute(self, ignore_cache=False):
         """
@@ -707,7 +708,8 @@ class Search(Request):
             es = get_connection(self._using)
 
             self._response = self._response_class(
-                self, es.search(index=self._index, body=self.to_dict(), **self._params)
+                self,
+                es.search(index=self._index, body=self.to_dict(), **self._params).body,
             )
         return self._response
 
@@ -799,7 +801,7 @@ class MultiSearch(Request):
             for s, r in zip(self._searches, responses["responses"]):
                 if r.get("error", False):
                     if raise_on_error:
-                        raise TransportError("N/A", r["error"]["type"], r["error"])
+                        raise ApiError("N/A", meta=responses.meta, body=r)
                     r = None
                 else:
                     r = Response(s, r)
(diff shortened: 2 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         run: |
           mkdir /tmp/elasticsearch
           wget -O - https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${{ matrix.es-version }}-linux-x86_64.tar.gz | tar xz --directory=/tmp/elasticsearch --strip-components=1
-          /tmp/elasticsearch/bin/elasticsearch -d
+          /tmp/elasticsearch/bin/elasticsearch -E xpack.security.enabled=false -E discovery.type=single-node -d
```
pquentin, line 79: I think disabling security is OK here as HTTPS is not a concern of this library: this is handled at the elasticsearch-py/elastic-transport-python level.

elasticsearch_dsl/search.py
```diff
@@ -693,7 +693,8 @@ def count(self):
 
         d = self.to_dict(count=True)
         # TODO: failed shards detection
-        return es.count(index=self._index, body=d, **self._params)["count"]
+        resp = es.count(index=self._index, query=d.get("query", None), **self._params)
```
pquentin, line 696: I did not keep `body` because the tests used `mocker.spy` and since body is dynamically added in elasticsearch-py the mock did not see it. But I could also use a different testing strategy if you prefer.

elasticsearch_dsl/search.py
```diff
@@ -799,7 +801,7 @@ def execute(self, ignore_cache=False, raise_on_error=True):
             for s, r in zip(self._searches, responses["responses"]):
                 if r.get("error", False):
                     if raise_on_error:
-                        raise TransportError("N/A", r["error"]["type"], r["error"])
+                        raise ApiError("N/A", meta=responses.meta, body=r)
```
pquentin, line 804: elasticsearch-py 8.x has two types of errors now, `ApiError` being used for errors reported by Elasticsearch itself and `TransportError` being used for errors at the transport layer (such as connection issues).

tests/conftest.py
```diff
@@ -131,8 +132,9 @@ def es_version(client):
 @fixture
 def write_client(client):
     yield client
-    client.indices.delete(index="test-*", ignore=404)
-    client.indices.delete_template(name="test-template", ignore=404)
+    for index_name in client.indices.get(index="test-*", expand_wildcards="all"):
+        client.indices.delete(index=index_name)
```
pquentin, line 136: Deleting indices with a wildcard require a cluster setting in Elasticsearch 8. I had to specify `expand_wildcards` as some tests are closing indices.

tests/conftest.py
```diff
@@ -160,55 +162,58 @@ def data_client(client):
...
-                    "_index": "test-index",
-                    "_type": "employee",
-                    "_id": "42",
-                    "_score": 11.123,
-                    "_routing": "elasticsearch",
-                    "_source": {
-                        "name": {"first": "Shay", "last": "Bannon"},
-                        "lang": "java",
-                        "twitter": "kimchy",
+    return ObjectApiResponse(
```
pquentin, line 165: In elasticsearch-py 8.x, the response pretends it is a dictionary to be backwards-compatible, but it does not support `__setitem__` that is needed for faceted search to work. This is why we explicitly request `.body` in `Search.execute`: this is an actual dictionary that supports `__setitem__`.

The alternative is to add `__setitem__` support in elastic-transport-python.

tests/test_connections.py
```diff
@@ -33,27 +38,36 @@ def test_default_connection_is_returned_by_default():
 
 
 def test_get_connection_created_connection_if_needed():
-    c = connections.Connections()
-    c.configure(default={"hosts": ["es.com"]}, local={"hosts": ["localhost"]})
+    c = connections.Connections(elasticsearch_class=DummyElasticsearch)
```
pquentin, line 41: Those tests are about `elasticsearch_dsl.connections`, not about how we store hosts in `Elasticsearch`. In elasticsearch-py 8.x, the hosts are technically available, but they're inside a deep hierarchy of classes, so using a dummy class avoids depending on those internals.

tests/test_connections.py
```diff
@@ -77,15 +94,16 @@ def test_remove_connection_removes_both_conn_and_conf():
...
 
     con = c.get_connection("testing")
-    assert [{"host": "es.com"}] == con.transport.hosts
+    assert con.hosts == ["https://es.com:9200"]
 
 
 def test_create_connection_adds_our_serializer():
-    c = connections.Connections()
-    c.create_connection("testing", hosts=["es.com"])
+    c = connections.Connections(elasticsearch_class=Elasticsearch)
```
pquentin, line 105: This test however uses the real client, for what it's worth.

tests/test_utils.py
```diff
@@ -93,7 +93,7 @@ class MyClass:
         def to_dict(self):
             return 42
 
-    assert serializer.serializer.dumps(MyClass()) == "42"
+    assert serializer.serializer.dumps(MyClass()) == b"42"
```
pquentin, line 96: In elasticsearch-py 8.x, serializers now explicitly encode the results to UTF-8. Which is more correct, after all serializing is supposed to output bytes.

pquentin (commented):
I've added a few comment to explain my thought processes when working on this pull request, including places where different decisions could have been made.

PR #1668, opened 2023-09-06
Fix Read the Docs build

pquentin:
The .readthedocs.yaml file is now required: https://blog.readthedocs.com/migrate-configuration-v2/

I faced two main issues here:
 * Our config file was not asking for sphinx-rtd-theme on Read the Docs, which was useful long ago but prevented the theme to be used today. This is why https://elasticsearch-dsl.readthedocs.io/en/latest/ shows the default theme instead (alabaster).
 * By default, Read the Docs installs old versions of sphinx and sphinx-rtd-theme, so I had to put lower bounds to avoid that.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,16 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3"
+
+python:
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - develop
+
+sphinx:
+  fail_on_warning: true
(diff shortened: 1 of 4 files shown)


JoshMock (approved):

JoshMock:
Looks like a couple small CI failures, but everything looks good on the whole. :+1:

pquentin:
I have enabled the Read the Docs pull request integration so the result can be seen here: https://elasticsearch-dsl--1668.org.readthedocs.build/en/1668/. Please take another look as I have continued to tweak things to make the build work.

JoshMock (approved):

PR #1671, opened 2023-09-07
Add 8.9.0 release to changelog on main

pquentin:


JoshMock (approved):
LGTM!

It might be beneficial to document all these steps to build/release this project, so we can store it in our team's docs for future reference.

PR #1689, opened 2023-12-22
Fix importing collapse from dict

miguelgrinberg:
Two minor bugs in the `collapse` implementation:

- The cloning of the `collapse` key was in `Request` instead of `Search`
- The `collapse` key was not importing from dict

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -120,7 +120,6 @@ class Request:
 
         self._doc_type = []
         self._doc_type_map = {}
-        self._collapse = {}
         if isinstance(doc_type, (tuple, list)):
             self._doc_type.extend(doc_type)
         elif isinstance(doc_type, collections.abc.Mapping):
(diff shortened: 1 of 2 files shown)


elasticsearch_dsl/search.py
```diff
@@ -294,7 +294,6 @@ def _clone(self):
         s = self.__class__(
             using=self._using, index=self._index, doc_type=self._doc_type
         )
-        s._collapse = self._collapse.copy()
```
pquentin, line 297: Great catch, thanks! Can you please move `self._collapse = {}` to `Search.__init__` too?

pquentin (changes requested):
Thank you! Looks great. One minor change left.

elasticsearch_dsl/search.py
miguelgrinberg: the `Search` class already initializes `self._collapse`. The one in `Request` should not be there, I'm removing it.

pquentin (approved):
Thanks! LGTM.

PR #1692, opened 2023-12-22
Add ``rank`` option to ``elasticsearch_dsl.search.Search``

miguelgrinberg:
This PR adds the `rank` option, to enable the use of the RRF algorithm to combine multiple sets of search results. Not sure if we want this published, given that `rank` is in technical preview at this point, but since I have implemented it for my own use I'm sharing the changes here.

### DIFF
--- a/elasticsearch_dsl/search.py
+++ b/elasticsearch_dsl/search.py
@@ -318,8 +318,9 @@ class Search(Request):
 
         self.aggs = AggsProxy(self)
         self._sort = []
-        self._collapse = {}
         self._knn = []
+        self._rank = {}
+        self._collapse = {}
         self._source = None
         self._highlight = {}
         self._highlight_opts = {}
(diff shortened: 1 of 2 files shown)


pquentin:
I like this and will want it when `knn` is merged. Maybe we should keep it undocumented since it's in technical preview?

miguelgrinberg:
@pquentin This is now ready. I left the docstring docs, including a note about RRF being in preview.

pquentin (approved):
Thanks! LGTM. I was able to use this in the "Hybrid Search using RRF" notebook:

```python3
import json

s = Search(using=client, index="book_index")
s = s[:5]
s = s.query("match", summary="python programming")
s = s.knn(
    field="title_vector",
    # generate embedding for query so it can be compared to `title_vector`
    query_vector=model.encode("python programming").tolist(),
    k=5,
    num_candidates=10,
)
s = s.rank(rrf={})

r = s.execute()
pretty_response(r)
```

PR #1714, opened 2024-03-08
async support

miguelgrinberg:
I attempted to process the `Search` class through `unasync`. WDYT?


pquentin (commented):
This is great!

* Focusing on Search first is a good idea
* I like that most methods are not doing I/O and can be in BaseSearch (which should be named `SearchBase` which is the convention in this project)
* Since this code is not generated like in elasticsearch-py, lint should check that `unasync(_async/search.py) == _sync/search.py`. The alternative is to not store `_sync/search.py` on disk and use unasync's setuptools integration.
* We need to use unasync on tests too, and the lint/docs builds need to work.
* Should we keep `search.py` so that `from elasticsearch_dsl.search import Search` still works? It's [unfortunately widely used](https://github.com/search?q=%22from+elasticsearch_dsl.search+import+Search%22&type=code)

miguelgrinberg:
> Since this code is not generated like in elasticsearch-py, lint should check that unasync(_async/search.py) == _sync/search.py. The alternative is to not store _sync/search.py on disk and use unasync's setuptools integration.

Do you mean to add this as a check, to make sure the code in `_sync` has not been hand edited? Like running unasync and then making sure `git status` is clean?

> We need to use unasync on tests too, and the lint/docs builds need to work.

Yes, I will look at that next.

> Should we keep search.py so that from elasticsearch_dsl.search import Search still works? It's [unfortunately widely used](https://github.com/search?q=%22from+elasticsearch_dsl.search+import+Search%22&type=code)

Right, I noticed that the search.py module is often referenced explicitly, so I thought it should be an alias for the generated `_sync/search.py`. We could also import the async classes in this module, so that people don't have to use `_async.search` in their imports.

miguelgrinberg:
@pquentin Making good progress on this. See the top issue comment. I will try to get through the document tests and see if that is a good stop point for this PR, depending on how much is left.

.github/workflows/ci.yml
```diff
@@ -61,7 +61,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [
-          "3.7",
```
pquentin, line 64: I'd like to do this in a separate pull request, as discussed offline, since it's going to many small changes that are best reviewed separately.

setup.cfg
```diff
@@ -11,3 +11,4 @@ filterwarnings =
     error
     ignore:Legacy index templates are deprecated in favor of composable templates.:elasticsearch.exceptions.ElasticsearchWarning
     ignore:datetime.datetime.utcfromtimestamp\(\) is deprecated and scheduled for removal in a future version..*:DeprecationWarning
+asyncio_mode = auto
```
pquentin, line 14: Please keep the default (strict mode) as this is more explicit and will help to add Trio support in the future.

setup.py
```diff
@@ -32,12 +32,20 @@
...
 develop_requires = [
+    "elasticsearch[async]",
+    "unasync",
     "pytest",
     "pytest-cov",
     "pytest-mock",
+    "pytest-asyncio",
     "pytz",
     "coverage",
+    "mock",  # needed to have AsyncMock in Python <= 3.7
```
pquentin, line 48: No longer needed as we plan to drop Python 3.7 support.

elasticsearch_dsl/document.py
```diff
@@ -15,474 +15,6 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
-import collections.abc
-from fnmatch import fnmatch
+from elasticsearch_dsl._sync.document import *  # noqa: F401, F403
```
pquentin, line 18: Can you please check if common tools such as mypy, pyright and jedi are able to autocomplete/type check correctly despite the star import?

noxfile.py
```diff
@@ -65,9 +65,10 @@ def format(session):
 
 @nox.session(python="3.12")
 def lint(session):
-    session.install("flake8", "black~=24.0", "isort")
+    session.install("flake8", "black~=24.0", "isort", "unasync", "setuptools")
     session.run("black", "--check", "--target-version=py37", *SOURCE_FILES)
     session.run("isort", "--check", *SOURCE_FILES)
+    session.run("python", "utils/run-unasync.py", "--check")
```
pquentin, line 71: Is this enough to make CI fail when the async/sync versions of a file are not synchronized?

pquentin (commented):
This looks good, thanks! I have reviewed the general approach so far, not yet the nitty gritty detail of what is sync, async or shared. But I'm not expecting any big issues there.

The next steps:

 * opening a pull request to drop Python 3.7 support (I can do that if you'd prefer)
 * add an async page to the docs, probably close to https://elasticsearch-py.readthedocs.io/en/v8.12.1/async.html (with general explanations around installation and usage and a complete API Reference at the end)
 * explain how to contribute code in CONTRIBUTING.rst

.github/workflows/ci.yml
miguelgrinberg: Yes, I'll create a separate PR. I removed it here temporarily to stop seeting the 3.7 failures.

setup.cfg
miguelgrinberg: I'm not sure how would that work with unasync. This is one example in which the code would need to be different. Async tests need a decorator to declare them as async, but that decorator needs to be removed in the sync conversion.

Example:

_async version:

```python
@pytest.mark.asyncio
async def test_some_asyncio_code():
    res = await library.do_something()
    assert b"expected result" == res
```

_sync version:

```python
def test_some_asyncio_code():
    res = library.do_something()
    assert b"expected result" == res
```

How do I tell unasync to remove the decorator here?

elasticsearch_dsl/document.py
miguelgrinberg: That is a good point. I'll check, but I suspect we'll have to do an explicit import here.

noxfile.py
miguelgrinberg: Yes. Here is an example output for the `lint` command after I added an `import re` in the _async version of document.py, without running unasync:

```
❯ nox -s lint
nox > Running session lint
nox > Creating virtual environment (virtualenv) using python3.12 in .nox/lint
nox > python -m pip install flake8 'black~=24.0' isort unasync setuptools
nox > black --check --target-version=py37 setup.py noxfile.py docs/ elasticsearch_dsl/ examples/ tests/ utils/
All done! ✨ 🍰 ✨
91 files would be left unchanged.
nox > isort --check setup.py noxfile.py docs/ elasticsearch_dsl/ examples/ tests/ utils/
Skipped 1 files
nox > python utils/run-unasync.py --check
reformatted /Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/elasticsearch_dsl/_sync_check/search.py
reformatted /Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/elasticsearch_dsl/_sync_check/document.py
reformatted /Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/elasticsearch_dsl/_sync_check/index.py

All done! ✨ 🍰 ✨
3 files reformatted, 1 file left unchanged.
diff -x __pycache__ elasticsearch_dsl/_sync/document.py elasticsearch_dsl/_sync_check/document.py
18a19
> import re
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/utils/run-unasync.py", line 101, in <module>
    main(check="--check" in sys.argv)
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/utils/run-unasync.py", line 88, in main
    subprocess.check_call(
  File "/Users/miguelgrinberg/.pyenv/versions/3.12.0/lib/python3.12/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['diff', '-x', '__pycache__', 'elasticsearch_dsl/_sync', 'elasticsearch_dsl/_sync_check']' returned non-zero exit status 1.
nox > Command python utils/run-unasync.py --check failed with exit code 1
nox > Session lint failed.
```

pquentin: Thanks!

elasticsearch_dsl/_async/search.py
```diff
@@ -0,0 +1,146 @@
...
+from ..utils import AttrDict
+
+
+class AsyncSearch(SearchBase):
+    def __aiter__(self):
+        """
+        Iterate over the hits.
+        """
+
+        class ResultsIterator:
```
miguelgrinberg, line 33: @pquentin Also would love your opinion on this iterator work. Due to using unasync, we're forced to have identical code for the async and sync sides, so I had to implement this iterator.

In reality there is no need for the iterator to be async even in the async side, since we retrieve all the results in a single call, but because the `execute()` call is async I didn't see any other choice other than making the whole iteration async.

Thoughts on this? I would have preferred to use a sync iterator for both async and sync (which I believe is what the main client does, right?), but I don't think we have a way to do that if we want the `AsyncSearch` object to be iterable.

pquentin: > In reality there is no need for the iterator to be async even in the async side, since we retrieve all the results in a single call, but because the `execute()` call is async I didn't see any other choice other than making the whole iteration async.

I can see why we would want a sync iterator and how making it async makes it look like we're maybe making multiple requests, yes. However, I think it's the current design of the library that forces our hand, not unasync. What we *could* do is encourage using `Response.__iter__` instead of `Search.__iter__` by documenting `for hit in Search(...).execute()` and `for hit in await AsyncSearch(...).execute()` instead of the current `for hit in Search(...)`. But on the other hand, `async for hit in AsyncSearch(...)` is not *that* bad?

Also, I think that allowing `Search.__iter__` but not `AsyncSearch.__iter__` would just be confusing. So in summary, I think `ResultsIterator` is fine.

> I would have preferred to use a sync iterator for both async and sync (which I believe is what the main client does, right?)

The main client does not expose iterators, except for helpers, which are async and sync.

miguelgrinberg: I think it looks nicer to iterate on the `AsyncSearch` and `Search` objects, without the explicit `execute()` call. So okay, I guess this is fine then.

pquentin: Yes, and using `async for` clarifies that I/O is in fact happening.

setup.py
miguelgrinberg: Removed

elasticsearch_dsl/document.py
miguelgrinberg: I have added explicit imports to document.py, search.py, update_by_query.py and mapping.py.

Question: Do we want to also import the async classes on these files? Because right now the async imports are all long and awkward:

```python
from elasticsearch_dsl._async.search import AsyncSearch
```

We could change this to:

```python
from elasticsearch_dsl.search import AsyncSearch
```

But of course, if we do this, we would always be import the sync and async versions into the Python environment, regardless of which one you want to use. Given that all the common methods are in the base class, this shouldn't add a lot of overhead, I think.

miguelgrinberg:
@pquentin Some more updates:

- Added update_by_query.py and mapping.py. The only remaining module that does I/O is analysis.py, but I'm not sure a conversion is necessary, because the only I/O happens in a `CustomAnalyzer` class that is not publicly exported. Do we want a `AsyncCustomAnalyzer` as well?
- Removed the star imports and added what I thought were sensible things to import in document.py, search.py, index.pt, update_by_query.py and mapping.py. I have an open question wrt also import the async classes into these files, so that they can be imported more easily.
- The faceted_search.py module references the `Search` class, but does not perform any I/O with it. Do we want to unasync it, or would it be better to pass the search class to use as an argument (defaulting to `Search`)?
- All tests that use the unasynced classes were also unasynced, including both regular and integration. It would probably be good for you to have a look and make sure I did not leave anything out. Coverage is at 93% (it is 94% for the currently released version, I think, so a very minor change that is likely caused by duplicating code that isn't covered via unasync).

Will work on the docs next!

elasticsearch_dsl/document.py
pquentin: As hinted above in our discussion last week, I think what should be encouraged is:

```python
from elasticsearch_dsl import AsyncSearch
```

We keep `search` only because it used to work but I would not encourage it:

```python
from elasticsearch_dsl.search import Search
```

And hopefully nobody will use this as `_async` starts as an underscore to show that it's private:

```python
from elasticsearch_dsl._async.search import AsyncSearch
```

pquentin:
> Added update_by_query.py and mapping.py. The only remaining module that does I/O is analysis.py, but I'm not sure a conversion is necessary, because the only I/O happens in a `CustomAnalyzer` class that is not publicly exported. Do we want a `AsyncCustomAnalyzer` as well?

There's some metaclass magic that I don't understand, but this is exposed through `analyzer()`, eg. see the `custom analyzer for names` in `examples/completion.py`. Does the example work when converted to asyncio? (We should consider leaving this for future work though, to not make this giant pull request even bigger.)

> The faceted_search.py module references the Search class, but does not perform any I/O with it. Do we want to unasync it, or would it be better to pass the search class to use as an argument (defaulting to Search)?

The second option seems better. Is there any disadvantage I'm missing?

> All tests that use the unasynced classes were also unasynced, including both regular and integration. It would probably be good for you to have a look and make sure I did not leave anything out. Coverage is at 93% (it is 94% for the currently released version, I think, so a very minor change that is likely caused by duplicating code that isn't covered via unasync). No tests were changed, added or removed.

Right, I do plan to read everything at least once, but I'm bound to miss things given the size of the pull request. Which is why I'm procrastinating :)

miguelgrinberg:
I have yet to convert the examples, so I'll test how analysis is used and determine if we need to do a conversion. None of the tests seemed to need an async analysis.py, but like you, I don't fully understand what's going on with this module.

For the faceted search class I agree with you that unasync is overkill. We just need to add the search class as an argument, and then we can create a `AsyncFacetedSearch` class that uses `AsyncSearch`.

miguelgrinberg:
I have asyncified the analisys.py as well. It wasn't clear at first, but there is metaclass magic between this module and the `DslBase` class that ends up performing I/O in the `simulate()` method of the analyzer.

miguelgrinberg:
@pquentin This PR is ready for review.

pquentin:
I'm going to review in the same order as you did your work, and the checklist will be in this comment.

- [ ] async connection management
- [ ] unasync Search
- [ ] check for invalid _sync directory in lint step
- [ ] unasync Search unit tests
- [ ] unasync Document
- [ ] unasync Search integration tests
- [ ] unasync Document tests
- [ ] unasync Index
- [ ] unasync Document integration tests
- [ ] unasync update_by_query
- [ ] unasync mappings
- [ ] unasync faceted_search
- [ ] unasync analyzer
- [ ] Examples
- [ ] Example integration tests
- [ ] Documentation updates

elasticsearch_dsl/search.py
```diff
@@ -15,924 +15,6 @@
...
-                        raise ApiError("N/A", meta=responses.meta, body=r)
-                    r = None
-                else:
-                    r = Response(s, r)
-                out.append(r)
-
-            self._response = out
-
-        return self._response
+from elasticsearch_dsl._async.search import AsyncMultiSearch, AsyncSearch  # noqa: F401
```
pquentin, line 18:
```suggestion
```

What I was trying to say in https://github.com/elastic/elasticsearch-dsl-py/pull/1714#discussion_r1531548995 is that we should *not* import the async classes here. Instead we should directly import `_async.search.AsyncSearch`  and `_async.search.AsyncMultiSearch` from the top-level `__init__.py`. My take was that all imports should be done at the top-level and we should not encourage users reaching out to subpackages as it limits our options going forward.

But maybe this is a lost cause and adding the async classes here helps users?

utils/run-unasync.py
```diff
@@ -0,0 +1,101 @@
...
+        "AsyncElasticsearch": "Elasticsearch",
+        "AsyncSearch": "Search",
+        "AsyncMultiSearch": "MultiSearch",
+        "AsyncDocument": "Document",
+        "AsyncIndexMeta": "IndexMeta",
+        "AsyncIndexTemplate": "IndexTemplate",
+        "AsyncIndex": "Index",
+        "AsyncUpdateByQuery": "UpdateByQuery",
+        "AsyncMapping": "Mapping",
+        "AsyncFacetedSearch": "FacetedSearch",
```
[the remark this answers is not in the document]
miguelgrinberg: My impression was that unasync did `AsyncX` to `SyncX`. Am I mistaken?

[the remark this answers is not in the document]
pquentin: Oh, you're right, sorry! Resolving.

utils/run-unasync.py
```diff
@@ -0,0 +1,101 @@
...
+        "tests/test_integration",
+        "tests/test_integration/test_examples",
+    ]
+    output_dir = "_sync" if not check else "_sync_check"
+
+    # Unasync all the generated async code
+    additional_replacements = {
+        "_async": "_sync",
+        "aiter": "iter",
+        "anext": "next",
```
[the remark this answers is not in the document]
miguelgrinberg: Yes, I actually had used them at some point, but then I ended up not using them. I'll remove these.

CONTRIBUTING.rst
```diff
@@ -25,22 +25,33 @@ The process for contributing to any of the Elasticsearch repositories is similar
    assure our users of the origin and continuing existence of the code. You only
    need to sign the CLA once.
 
-2. Run the test suite to ensure your changes do not break existing code:
+2. Many classes included in this library are offered in two versions, for
+   asynchronous and synchronous Python. When working with these classes, you only
+   need to make changes to the asynchronous code, located in the *_async*
+   subdirectory of the source tree. Once you've made your changes, run the
```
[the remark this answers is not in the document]
miguelgrinberg: Yes, good point.

tests/conftest.py
```diff
@@ -16,29 +16,34 @@
 #  under the License.
 
 
+import asyncio
```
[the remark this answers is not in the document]
miguelgrinberg: Some of these are not identical, so they cannot be managed by unasync. Maybe what makes more sense is to try to base one of the fixtures into the other, like I did for the `async_data_client`:

```python
@pytest_asyncio.fixture
async def async_data_client(data_client, async_client):
    yield async_client
```

I can probably do some of the others in this way, which is actually cleaner, as the logic is in one place.

elasticsearch_dsl/search.py
miguelgrinberg: I'm fine either way. I added them here because some of the tests imported this module. I'm not sure how common this is, but someone who wants to use both `Search` and `Multisearch` or both `AsyncSearch` and `AsyncMultiSearch` may want to only import `search` and access the classes from there. Is there anything wrong with that? If we do not import the async classes then the equivalent import is ugly:

```python
from elasticsearch_dsl._async import search
```

pquentin: How is `from elasticsearch_dsl.search import Search, MultiSearch` better than `from elasticsearch_dsl import Search, MultiSearch`? It's not like it would reduce the import time, right?

tests/conftest.py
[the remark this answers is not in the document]
pquentin: Sounds great to me.

elasticsearch_dsl/search.py
miguelgrinberg: Sorry, I wasn't very clear. What I'm saying is that you may want to do this:

```python
from elasticsearch_dsl import search

s = search.Search()
ms = search.MultiSearch()
```

or the async version:

```python
from elasticsearch_dsl import search

s = search.AsyncSearch()
ms = search.AsyncMultiSearch()
```

If we do not include the async classes in `search`, then the only other way to do this is with the `_async` module, which is not as nice:

```python
from elasticsearch_dsl._async import search

s = search.AsyncSearch()
ms = search.AsyncMultiSearch()
```

pquentin: I certainly don't want to encourage importing `_async`. I have seen zero evidence of users [doing what you suggest though](https://github.com/search?q=%22from+elasticsearch_dsl+import+search%22&type=code&ref=advsearch&p=1). While there's [plenty of evidence of importing through `search` for no reasons](https://github.com/search?q=%22from+elasticsearch_dsl.search+import+Search%22&type=code&ref=advsearch).

I'll let you make the decision here.

miguelgrinberg: I don't know if users do it or not, but we do it all over the place in tests:

https://github.com/elastic/elasticsearch-dsl-py/blob/main/tests/test_search.py#L22
https://github.com/elastic/elasticsearch-dsl-py/blob/main/tests/test_document.py#L32
https://github.com/elastic/elasticsearch-dsl-py/blob/main/tests/test_mapping.py#L20
https://github.com/elastic/elasticsearch-dsl-py/blob/main/tests/test_integration/test_mapping.py#L20

And I have to agree that if you are going to use a bunch of things from a module, it sometimes is more convenient to import the module instead of all the individual things.

elasticsearch_dsl/_async/index.py
```diff
@@ -0,0 +1,540 @@
...
+        if self._name is None:
+            raise ValueError("You cannot perform API calls on the default index.")
+        return get_connection(using or self._using)
+
+    connection = property(_get_connection)
+
+    def get_or_create_mapping(self):
+        if self._mapping is None:
+            self._mapping = AsyncMapping()
+        return self._mapping
```
pquentin, line 69: I was confused for a bit because this sync method in `AsyncIndex` is only called by `IndexBase`.

Is there value in making `IndexBase` (and other bases) as inheriting from [`abc.ABC`](https://docs.python.org/3/library/abc.html) to avoid direct instantiation and declaring sync methods that are in both async/sync child classes as abstract methods in `IndexBase`? That way `get_or_create_mapping` *will* show up in `IndexBase`, but will stay implemented in `AsyncIndex` and `Index`. The same applies to `as_template`.

Would it be useful to do for async methods too? Not sure. Especially as the base will have to be either async or sync, which will require type ignores in the child classes when we add mypy.

tests/_sync/test_index.py
```diff
@@ -68,7 +68,7 @@ def test_cloned_index_has_copied_settings_and_using():
 
 def test_cloned_index_has_analysis_attribute():
     """
-    Regression test for Issue #582 in which `Index.clone()` was not copying
+    Regression test for Issue #582 in which `AsyncIndex.clone()` was not copying
```
pquentin, line 71: This is fine but funny :)

pquentin (commented):
I've reviewed a third of the files now.

tests/_async/test_document.py
```diff
@@ -0,0 +1,638 @@
...
+        await md.delete(using="mock")
+
+
+async def test_update_no_fields():
+    md = MyDoc()
+    with raises(IllegalOperation):
+        await md.update()
+
+
+async def test_search_with_custom_alias_and_index(async_mock_client):
```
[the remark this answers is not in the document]
miguelgrinberg: I don't understand this one. Why is `async` wrong here?

tests/_async/test_search.py
```diff
@@ -0,0 +1,687 @@
...
+
+
+async def test_iter_iterates_over_hits():
+    s = search.AsyncSearch()
+    s._response = [1, 2, 3]
+
+    r = []
+    async for hit in s:
+        r.append(hit)
+    assert [1, 2, 3] == r
```
pquentin, line 57:
```suggestion
    assert [1, 2, 3] == [hit async for hit in s]
```

elasticsearch_dsl/_async/index.py
miguelgrinberg: The `get_or_create_mapping()` method is also called by the `Index` and `AsyncIndex` classes, in the `load_mapping()` method:

```python
async def load_mappings(self, using=None):
        await self.get_or_create_mapping().update_from_es(
            self._name, using=using or self._using
        )
```

But given that this method is sync in both interfaces, I think for this one we should move the method to the base class, then add a `get_mapping_class()` in the subclasses each returning its variant of the mapping class. You think?

miguelgrinberg: And also, maybe we should make all these base classes abstract, but honestly I have never worried about that in other projects of my own, it always seemed like have a `Base` prefix or suffix was sufficient for people to know this isn't a class meant to be instantiated.

pquentin: Sure, that works, and is better, but it only moves the problem. I continue to think that:

 * making the Base classes inheriting from `abc.ABC` is useful, at least to prevent direct instantiation
 * declaring sync methods from child classes (`get_mapping_class()` now) is probably helpful?

pquentin (commented):
I've reviewed another third of the files: docs, examples and their tests are left.

elasticsearch_dsl/_async/index.py
pquentin: Our comments crossed. I'm happy to leave out `ABC` and abstract methods if you don't think they help.

tests/_async/test_mapping.py
```diff
@@ -0,0 +1,223 @@
...
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+import json
+
+from elasticsearch_dsl import Keyword, Nested, Text, analysis
+from elasticsearch_dsl._async import mapping
```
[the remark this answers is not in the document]
miguelgrinberg: Yes, there's a few. I used a search/replace approach to convert the code, so I kept these as close as the originals as I could. But I agree, I'll remove these, because if we use them here people will copy them and that's not good.

docs/asyncio.rst
```diff
@@ -0,0 +1,92 @@
+.. _asyncio:
+
+Using Asyncio with Elasticsearch DSL
+====================================
+
+The ``elasticsearch-dsl`` package supports async/await with `Asyncio <https://docs.python.org/3/library/asyncio.html>`__.
```
pquentin, line 6: Please lowercase `asyncio`

docs/asyncio.rst
```diff
@@ -0,0 +1,92 @@
...
+
+       responses = await ms.execute()
+
+       for response in responses:
+           print("Results for query %r." % response.search.query)
+           for hit in response:
+               print(hit.title)
+
+Asynchronous Documents, Indexes, and more
+-----------------------------------------
```
pquentin, line 81: Should we mention the one real difference, async_simulate vs. simualte?

docs/api.rst
```diff
@@ -11,39 +11,70 @@ Search
...
+   :inherited-members:
+   :members:
+
+.. autoclass:: AsyncSearch
+   :inherited-members:
+   :members:
+
+.. autoclass:: AsyncMultiSearch
+   :inherited-members:
    :members:
```
pquentin, line 27: I believe the async API should be distinct from the sync API. I quite like the elasticsearch-py away, which is just to include the API at the bottom the page. But it could also be a different page.

pquentin (commented):
I've reviewed everything now! Thanks again for this massive body of work.

## Examples

I believe examples should be using unasync too. I'm already seeing divergences (such as `yield from`, moving away from `@property` or closing connections) that should not be here, and they will only grow in the future. This would require:

1. having a script (run-unasync.py or a wrapper) to replace `asyncio.run(main())` with `main()` and remove `import asyncio`, since unasync can't do it itself.
1. checking we can generate sync examples from async example just like the rest of the files with CI

## Docs

Regarding documentation, we should mention async/await support in the introduction and have an async example in index.rst

tests/_async/test_mapping.py
```diff
@@ -17,12 +17,11 @@
 
 import json
 
-from elasticsearch_dsl import Keyword, Nested, Text, analysis
-from elasticsearch_dsl._async import mapping
+from elasticsearch_dsl import AsyncMapping, Keyword, Nested, Text, analysis
```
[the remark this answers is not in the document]
miguelgrinberg: I could have done that, but if we are saying we prefer direct imports I thought it would be better to switch. I still think it doesn't hurt to have the async imports in `mapping`, `search`, etc. if anything for symmetry, since obviously we cannot remove the sync classes from there.

docs/asyncio.rst
miguelgrinberg: There's a buch of uppercase Asyncio references in the main client. This is a copy/paste from there.

docs/asyncio.rst
miguelgrinberg: I think we should, yes.

tests/_async/test_mapping.py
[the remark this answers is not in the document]
pquentin: No real reason, I guess I just felt bad about you doing more work than necessary. Resolving.

docs/asyncio.rst
pquentin: https://github.com/elastic/elasticsearch-py/pull/2516 :)

miguelgrinberg:
@pquentin I think I have addressed everything except the unasync for the examples. Please have a look at the new structure for the docs with a separate reference page for the async classes.

I will look at the examples next week. I would like to avoid having to change the locations of the sync examples, I'm going to need a special unasync rule to handle going from `examples/async/x.py` to `examples/x.py`. I don't think we want the ugly `_sync` and `_async` for these.

pquentin:
The [current navigation](https://elasticsearch-dsl--1714.org.readthedocs.build/en/1714/index.html) feels off:

image

We have two pages about asyncio, one with text and the other with the reference API. And the async reference API has a different name from the sync reference API. Maybe we should have the following navigation instead?

* Tutorials
  * Quickstart (with the current content of the index page?)
* How-to guides
  * Search DSL
  * Persistence
  * Faceted Search
  * Update by Query
  * Using asyncio with Elasticsearch DSL
* Reference
  * API Documentation
  * Async API Documentation

The section names come from https://diataxis.fr/. Here's an example of Sphinx documentation using sections in the navigation: https://esrally.readthedocs.io/en/stable/. Since the links don't reference the section, I believe all existing links will continue to work.

> Regarding documentation, we should mention async/await support in the introduction and have an async example in index.rst

I don't believe this was addressed yet.

utils/run-unasync.py
```diff
@@ -75,24 +89,37 @@ def main(check=False):
...
-
-    if check:
-        # make sure there are no differences between _sync and _sync_check
-        for source_dir in source_dirs:
+    for dir in source_dirs:
+        output_dir = f"{dir[0]}_sync_check/" if check else dir[1]
+        subprocess.check_call(["black", "--target-version=py38", output_dir])
+        subprocess.check_call(["isort", output_dir])
+        for file in glob("*.py", root_dir=dir[0]):
+            # remove asyncio from sync files
```
pquentin, line 97: I have a few doubts about 1/ doing this for all files and not only examples and 2/ using a sed command carefully crafted to work on both Linux and macOS. This seems fragile, but it works, so let's do it.

pquentin (approved):
The new doc navigation is amazing, thanks! And thanks for working on the examples too.

Looks good to me.

utils/run-unasync.py
miguelgrinberg: Yes, 100% agree on the fragility. I don't feel doing this for all files is a problem, since what we are doing is the correct thing to do as a general case. I could actually make the regex in the sed a bit smarter and capture the function name that is passed to `asyncio.run()` instead of having it hardcoded. I would think this type of transformation would make sense to add to unasync (although not with sed!).

PR #1717, opened 2024-03-19
Remove support for Python 3.7

miguelgrinberg:
As discussed, we are removing Python 3.7 from the actively supported versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [
-          "3.7",
           "3.8",
           "3.9",
           "3.10",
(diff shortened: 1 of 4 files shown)


pquentin (approved):
Thanks! LGTM. I thought this would be more involved than that.

Can you please run `pip install pyupgrade && pyupgrade --py38-plus **/*.py`? This will update build-dist.py to use `shlex.join` (new in Python 3.8) and `python -m build` (as invoking setup.py is deprecated).

You also need to change `py37` to `py38` in `noxfile.py` (and make sure to do it in `utils/run-unasync.py` in the async PR). Not that it makes any difference.

Also, yay, we can now use the walrus operator, positional-only parameters and use = in f-strings! https://docs.python.org/3/whatsnew/3.8.html

PR #1771, opened 2024-04-09
Added support for slicing multiple times in Search class

### LINKED ISSUE #799: Slicing twice doesn't work
Iterators are expected to allow slicing twice, i.e.

```python
>>> list(range(0,100)[10:15])
[10, 11, 12, 13, 14]
>>> list(range(0,100)[10:][:5])
[10, 11, 12, 13, 14]
```

This means essentially that `[a:a+b]` equals `[a:][:b]`

With a `Search` object slicing twice doesn't work:
 
```python
>>> Search()[10:15].to_dict()
{'from': 10, 'query': {'match_all': {}}, 'size': 5}
>>> Search()[10:][:5].to_dict()
{'from': 0, 'query': {'match_all': {}}, 'size': 5}
```

(Sorry for submitting an incomplete issue first, I accidentally hit the wrong key)

miguelgrinberg:
Fixes #799

This PR implements Pythonic slicing for the `Search` object.

Note: due to introducing backward incompatibilities, this change is scheduled to be incorporated in release 8.15. Starting with release 8.13.1 a warning will alert developers slicing results multiple times that this breaking change is coming.


elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,27 @@ def __getitem__(self, n):
...
-            s._extra["from"] = n
-            s._extra["size"] = 1
-            return s
+            slice_start = n
+            slice_stop = n + 1
+        # Elasticsearch won't get all results so we default to size: 10 if
+        # stop not given.
+        old_from = s._extra.get("from", 0)
+        old_to = old_from + s._extra.get(
+            "size", slice_stop or (slice_start or old_from) + 10
```
pquentin, line 367: Why would `old_to` depend on `slice_stop` or `slice_start` which are the new values?

elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,27 @@ def __getitem__(self, n):
...
+        )
+        new_from = old_from + (slice_start or 0)
+        new_to = (
+            min(old_to, old_from + slice_stop) if slice_stop is not None else old_to
+        )
```
pquentin, line 372: I know the previous code did that, but can we please avoid a ternary expression here to make the code less dense?

elasticsearch_dsl/search_base.py
pquentin, line 367: Why aren't the parens like `(slice_stop or slice_start or old_from)`?

pquentin (commented):
Thanks! The tests make sense, but the code is difficult to read for me, so I asked clarification questions.

elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,27 @@ def __getitem__(self, n):
             # If negative slicing, abort.
             if n.start and n.start < 0 or n.stop and n.stop < 0:
                 raise ValueError("Search does not support negative slicing.")
```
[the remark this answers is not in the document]
miguelgrinberg: Hmm, no. Neither this or the next error conditions are tested. I'll add tests for these.

elasticsearch_dsl/search_base.py
miguelgrinberg: I undesrtand that this looks wrong, but I think it is the correct approach for us. The problem is that the initial default of size=10 is a "soft" default that needs to be overriden if the first slicing operation specifies a different size.

Consider this example:

- we start with a default slice of `[0:10]`.
- now you ask for `Search()[:40]`. With a normal slicing operation, you would get `[0:10]` again. What we want is for this first slicing operation to change the size.

I hope this makes sense!

miguelgrinberg: There was a minor mistake in this logic. I have now expanded this logic into a set of conditionals to make everything more clear. It was incorrect to abuse the `or` operator here, since `slice_stop` and `slice_start` can be `None` or `0` and these should be interpreted differently.

elasticsearch_dsl/search_base.py
```diff
@@ -363,13 +363,22 @@ def __getitem__(self, n):
         # Elasticsearch won't get all results so we default to size: 10 if
         # stop not given.
```
pquentin, line 364: This comment can be removed now as it has been duplicated in the `slice_start is not None` case

elasticsearch_dsl/search_base.py
pquentin: > now you ask for `Search()[:40]`. With a normal slicing operation, you would get `[0:10]` again. What we want is for this first slicing operation to change the size.

I don't know, it's not that intuitive. I assumed `s[:10][:40]` was the same as `s[:10]`. Can we add a test about this?

The Django version linked by Honza in the original issue is much simpler: https://github.com/django/django/blob/ca5cd3e3e8e53f15e68ccd727ec8fe719cc48099/django/db/models/sql/query.py#L2116-L2137. Decoupling high/low and from/size and the default size limit sounds smart.

miguelgrinberg:
> I don't know, it's not that intuitive. I assumed s[:10][:40] was the same as s[:10]. Can we add a test about this?

Yes, this is actually how it works. Just added a couple more tests around this.

elasticsearch_dsl/search_base.py
miguelgrinberg: > I don't know, it's not that intuitive. I assumed s[:10][:40] was the same as s[:10]. Can we add a test about this?

Yes, this is actually how it works. Just added a couple more tests around this.

The idea here is that `s[:10][:40]` == `s[:10]`, but `s[:40]` should get 40 results, not 10.

The Django code is simpler, I think, because it does not have an implied default of 10 results, which we need to keep for backwards compatibility.

miguelgrinberg:
@pquentin I have simplified the slicing even more. If the from or the size aren't specified in the slicing, then they are not included in the final request body. So for example:
- `[4:]` renders to `{"from": 4}`
- `[:4]` renders to `{"size": 4}`
- `[2:4]` renders to `{"from": 2, "size": 2}`
- `[:]` renders to `{}`

elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,34 @@ def __getitem__(self, n):
...
             if n < 0:
                 raise ValueError("Search does not support negative indexing.")
-            s._extra["from"] = n
-            s._extra["size"] = 1
-            return s
+            slice_start = n
+            slice_stop = n + 1
+
+        old_from = s._extra.get("from")
+        old_to = (old_from or 0) + s._extra["size"] if "size" in s._extra else None
```
pquentin, line 365: I initially read this as `old_to = (old_from or 0) + (s._extra["size"] if "size" in s._extra else None)`. We could use parents to make this more obvious, but arguable the expression is complex enough to *not* use a ternary expressions?

```suggestion
        old_to = None
        if "size" in s._extra:
            old_to = (old_from or 0) + s._extra["size"]
```

You can even use the walrus operator if you feel fancy:

```suggestion
        old_to = None
        if old_size := s._extra.get(size):
            old_to = (old_from or 0) + old_size
```

elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,34 @@ def __getitem__(self, n):
...
+        if slice_start is not None:
+            new_from = (old_from or 0) + slice_start
+        else:
+            new_from = old_from
+        if slice_stop is not None:
+            new_to = (old_from or 0) + slice_stop
+            if old_to is not None and old_to < new_to:
+                new_to = old_to
+        else:
+            new_to = old_to
```
pquentin, line 376: Using the pattern above, this can probably be simplified too:

```suggestion
        new_from = old_from
        if slice_start is not None:
            new_from = (old_from or 0) + slice_start

        new_to = old_to
        if slice_stop is not None:
            new_to = (old_from or 0) + slice_stop
            if old_to is not None and old_to < new_to:
                new_to = old_to
```

elasticsearch_dsl/search_base.py
```diff
@@ -352,20 +352,34 @@ def __getitem__(self, n):
...
+
+        if new_from is not None:
+            s._extra["from"] = new_from
+        if new_to is not None:
+            s._extra["size"] = max(0, new_to - (new_from or 0))
```
pquentin, line 381: Would it help to store `from` and `to` instead of `from` and `size`?

pquentin (commented):
Thanks! The logic looks good. It just needs documentation.

My only remaining worry is that this is a breaking change. Would it be useful to warn about using multiple slices in 8.13.1 and 8.14 and then switch to the new behavior in 8.15? Some users do pay attention to our warnings: https://github.com/elastic/elasticsearch-py/issues/2518#issuecomment-2036207479.

elasticsearch_dsl/search_base.py
miguelgrinberg: I don't think so, because we would still need to support people doing `search.extra(size=50)` to not break backwards compatibility.

pquentin (commented):
Thanks! LGTM. We'll have to make sure to label this as a breaking change in the release notes.

pquentin (approved):

PR #1772, opened 2024-04-10
Better coverage report after tests

miguelgrinberg:
Expand the coverage report printed after the test run to include missing lines and branches. Also generate an HTML report for reviewing coverage locally.


pquentin (approved):
Thanks! Should we upload the HTML coverage as a GitHub artifact? (Don't do it for me, I don't use HTML coverage reports.)

miguelgrinberg:
We add them if anybody wants to see them, but I just look at them locally, that seems sufficient for me.

PR #1773, opened 2024-04-10
Corrected documentation of sort option

### LINKED ISSUE #1318: FacetedSearch does not accept `sort` configuration
- ElasticSearch: 7.5.2 on Ubuntu
- Python: 3.8.1
- elasticsearch-dsl: 7.1.0

[The document](https://github.com/elastic/elasticsearch-dsl-py/blob/43c71ce6a22ed8ac0e44be6fa27dd680375a8b0a/docs/faceted_search.rst#configuration) says FaucetedSearch accept configuration option `sort` as a class attribute.

> `sort`
>   tuple or list of fields on which the results should be sorted. 

So I tried to create the following class.
```python
class BlogSearch(FacetedSearch):
    index = 'blogs'

    doc_types = [Blog, Post]
    fields = ['title', 'content']

    sort = ('-published_at',)

    def search(self):
        s = super().search()
        return s.filter('term', published=True)
```

Then it raises TypeError.

	>>> BlogSearch().execute()
  	File "<console>", line 1, in <module>
  	File "***/site-packages/elasticsearch_dsl/faceted_search.py", line 271, in __init__
    	self._s = self.build_search()
  	File "***/site-packages/elasticsearch_dsl/faceted_search.py", line 381, in build_search
    	s = self.sort(s)
	TypeError: 'tuple' object is not callable

I tried `sort = '-publisehd_at'` but got the same error. What kind of argument should be set?

`BlogSearch(sort=('-published_at',))` works fine, but I'd like to set the default sort order.

miguelgrinberg:
Fixes #1318


pquentin (approved):
Thanks! LGTM.

Issue #1451, opened 2020-11-13
__and__ operator fails for Bool with minimum_should_match in percentage


mmoriniere:
Hello,

I ran into on an error when trying to chain two such queries :
```python
MyDocument.search().query(
    Bool(filter=Term(disabled=False))
).query(
    Bool(should=[my_should_clauses], minimum_should_match="-10%")
)
```
(where MyDocument derives from the Document class of django_elasticsearch_dsl).

The error happens [here](
https://github.com/elastic/elasticsearch-dsl-py/blob/8a9c374a59928abdbc6b1117d54852bda021c4ad/elasticsearch_dsl/query.py#L213), and there actually is a TODO three lines above suggesting that this problem is known and ought to be fixed.

Is there any plan to support such a feature ?

Note: I already found two workarounds to avoid this issue :
1. Refactor my code to use just one query,
2. Make the second query look like this `Bool(filter=Bool(should=[my_should_clauses], minimum_should_match="-10%"))`.

miguelgrinberg:
A more explicit error message is now shown, indicating that only positive integers are currently supported for `minimum_should_match`.

PR #1774, opened 2024-04-11
Added error checking for unsupported minimum_should_match values

miguelgrinberg:
This is a small improvement regarding #1451, so that the error message provides guidance on what's wrong. The developer can then decide if it makes sense to rewrite the terms using a positive minimum_should_match, or create a combined expression manually.


pquentin (approved):
Thanks! LGTM.

PR #1777, opened 2024-04-12
Added inner_hits option to kNN search

miguelgrinberg:
I've missed the `inner_hits` option for kNN, needed to retrieve nested documents.


pquentin (approved):
Thanks! LGTM.

PR #1778, opened 2024-04-12
Vector search example

miguelgrinberg:
Depends on #1776 and #1777.

### DIFF
--- a/examples/async/vectors.py
+++ b/examples/async/vectors.py
@@ -20,21 +20,27 @@
 
 Requirements:
 
-$ pip install nltk sentence_transformers tqdm elasticsearch_dsl
+$ pip install nltk sentence_transformers tqdm elasticsearch-dsl[async]
 
 To run the example:
 
 $ python vectors.py "text to search"
 
-The index will be created automatically if it does not exist. Add `--create` to
-regenerate it.
+The index will be created automatically if it does not exist. Add
+`--recreate-index` to regenerate it.
 
-The example dataset includes a selection of workplace documentation. The
-following are good example queries to try out:
+The example dataset includes a selection of workplace documents. The
+following are good example queries to try out with this dataset:
 
 $ python vectors.py "work from home"
 $ python vectors.py "vacation time"
-$ python vectors.py "bring a bird to work"
+$ python vectors.py "can I bring a bird to work?"
+
+When the index is created, the documents are split into short passages, and for
+each passage an embedding is generated using the open source
+"all-MiniLM-L6-v2" model. The documents that are returned as search results are
+those that have the highest scored passages. Add `--show-inner-hits` to the
+command to see individual passage results as well.
 """
 
 import argparse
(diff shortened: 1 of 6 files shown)


examples/async/vectors.py
```diff
@@ -0,0 +1,170 @@
...
+    model = WorkplaceDoc.get_embedding_model()
+    search = WorkplaceDoc.search().knn(
+        field="passages.embedding",
+        k=5,
+        num_candidates=50,
+        query_vector=list(model.encode(query)),
+        inner_hits={"size": 3},
+    )
+    async for hit in search:
+        print(f"Document: {hit.name} (Category: {hit.category}")
```
pquentin, line 139:
```suggestion
        print(f"Document: {hit.name} (Category: {hit.category})")
```

examples/async/vectors.py
```diff
@@ -0,0 +1,170 @@
...
+        for passage in hit.meta.inner_hits.passages:
+            print(f"  - [Score: {passage.meta.score}] {passage.content!r}")
```
pquentin, line 141: I'm not sure the !r is that useful here.

examples/async/vectors.py
```diff
@@ -0,0 +1,170 @@
...
+    return parser.parse_args()
+
+
+async def main():
+    args = parse_args()
+
+    # initiate the default connection to elasticsearch
+    async_connections.create_connection(hosts=[os.environ["ELASTICSEARCH_URL"]])
+
+    if args.create or not await WorkplaceDoc._index.exists():
```
pquentin, line 160: Not sure what the point of create is if we're going to create anyway? Is that meant to "recreate"?

pquentin (commented):
This is great, thanks!

examples/async/vectors.py
miguelgrinberg: The method I'm using to split into sentences sometimes leaves a `\n` in some of the parts. With the `!r` those are rendered as `\n` instead of destroying the small attempt I've made of having a neat layout.

examples/async/vectors.py
miguelgrinberg: Yes. I should rename this `--recreate` for more clarity.

examples/async/vectors.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+async def create():
+
+    # create the index
+    await WorkplaceDoc._index.delete(ignore_unavailable=True)
+    await WorkplaceDoc.init()
+
+    # download the data
+    dataset = json.loads(urlopen(DATASET_URL).read())
```
miguelgrinberg, line 113: @pquentin any suggestions on how to read this URL in a way that unasync can do the right thing? Technically this is not something we'd want to promote for the async example.

pquentin (approved):
Thanks! LGTM. Just need to skip on 8.0.

examples/async/vectors.py
```diff
@@ -128,24 +134,24 @@ async def create():
...
         num_candidates=50,
         query_vector=list(model.encode(query)),
-        inner_hits={"size": 3},
+        inner_hits={"size": 2},
     )
-    async for hit in search:
-        print(f"Document: {hit.name} (Category: {hit.category}")
-        for passage in hit.meta.inner_hits.passages:
-            print(f"  - [Score: {passage.meta.score}] {passage.content!r}")
-        print("")
```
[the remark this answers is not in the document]
miguelgrinberg: Amen.

PR #1782, opened 2024-04-18
Switch to pytest-asyncio's strict mode for async tests

miguelgrinberg:
The trick is to using the following for async tests:

    @pytest.mark.asyncio

which is convert to the following for the unasynced tests:

    @pytest.mark.sync

This eliminates issues with unused imports, which would need to be taken care of in some, but not all files. WDYT?

### DIFF
--- a/utils/run-unasync.py
+++ b/utils/run-unasync.py
@@ -70,6 +70,7 @@ def main(check=False):
         "async_pull_request": "pull_request",
         "async_examples": "examples",
         "assert_awaited_once_with": "assert_called_once_with",
+        "pytest_asyncio": "pytest",
     }
     rules = [
         unasync.Rule(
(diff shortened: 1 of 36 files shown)


pquentin (commented):
Thanks!

I think `syncio` is just too close to `asyncio` (to the point where my phone autocorrects it back to asyncio). I would be OK with `sync` or any other word with a Levenshtein distance >= 2 from `asyncio`.

(I would also be OK with a sed solution until unasync learns to deal with multi-token deletion rules.)

utils/run-unasync.py
```diff
@@ -70,6 +70,7 @@ def main(check=False):
         "async_pull_request": "pull_request",
         "async_examples": "examples",
         "assert_awaited_once_with": "assert_called_once_with",
+        "pytest_asyncio": "pytest",
```
pquentin, line 73: Why is this needed now? Also, I think `@pytest_asyncio.fixture` and `@pytest.fixture` are probably clearer than `@fixture`, but that can be addressed further down the road.

pquentin (approved):
Thanks! LGTM.

utils/run-unasync.py
miguelgrinberg: This is because pytest-asyncio's auto mode also takes care of reconfiguring any async fixtures declared with `@pytest.fixture` appropriately. In strict mode you have to use `@pytest_asyncio.fixture` instead.

I agree that `@pytest.fixture` is better than `@fixture` here, I'll change it.

PR #1783, opened 2024-04-23
explicit use of fixture decorator

miguelgrinberg:
This was one of the last suggestions in #1782, to use explicit `@pytest.fixture` and `@pytest_asyncio.fixture` decorators. I inadvertently left this change out of that PR, so here it is.


pquentin (approved):
Thanks! LGTM.

PR #1786, opened 2024-04-25
Configure a unique user agent for usage tracking purposes

miguelgrinberg:
Issues/questions
- Attempting to inject a user agent in a user provided client feels not very robust. I'm looking for an `options` method before calling it in case they give us a wrapper and not an actual client, but it can still break.
- Mock clients need a bit of extra handling, because setting the user agent requires the client to be copied:
        ```
        client = client.options(headers={"user-agent": "..."})
        ```
This causes the original mock to not get any of the calls, unless the mock is configured with an `options()` method. Should we modify clients in place, instead of making a copy?

### DIFF
--- a/elasticsearch_dsl/connections.py
+++ b/elasticsearch_dsl/connections.py
@@ -58,7 +58,7 @@ class Connections:
         """
         Add a connection object, it will be passed through as-is.
         """
-        self._conns[alias] = conn
+        self._conns[alias] = self._with_user_agent(conn)
 
     def remove_connection(self, alias):
         """
@@ -82,7 +82,7 @@ class Connections:
         """
         kwargs.setdefault("serializer", serializer)
         conn = self._conns[alias] = self.elasticsearch_class(**kwargs)
-        return conn
+        return self._with_user_agent(conn)
 
     def get_connection(self, alias="default"):
         """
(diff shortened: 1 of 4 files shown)


pquentin (commented):
> * Attempting to inject a user agent in a user provided client feels not very robust. I'm looking for an `options` method before calling it in case they give us a wrapper and not an actual client, but it can still break.

This specific bit sounds OK to me. If you don't have an options() function, then it's already going to break I think?

> * Mock clients need a bit of extra handling, because setting the user agent requires the client to be copied:
>   `client = client.options(headers={"user-agent": "..."})`
>  This causes the original mock to not get any of the calls, unless the mock is configured with an `options()` method. Should we modify clients in place, instead of making a copy?

This bit actually worries me more, because we're going to break third-party tests that also use mocking. I looked into modifying the clients in place, but that's bit more complicated than it sounds, because we `freeze()` the headers early:

https://github.com/elastic/elasticsearch-py/blob/8511ba7e033adb39058c078bf75562f34d243288/elasticsearch/_sync/client/utils.py#L135-L136

Maybe there's a way to *replace* the `_headers` completely. We also pass the headers to the transport which does its own handling so that would have to work. Using `.options()` sounds much easier so I'm OK to merge this for 8.14 as long as we mention it in the release notes.

Also, can we test that user agent is in fact set to `elasticsearch-dsl-py/{__versionstr__}`?

miguelgrinberg:
Closing. Will open a new PR with a different approach.

PR #1787, opened 2024-04-26
Fixed documentation example way to access search object

### LINKED ISSUE #1223: Search field not accessible in Response object for MultiSearch
The docs say that you should be able to access the search object in the response, but after inspecting the code and trying in code, I get an attribute error. I believe it's only available at `._search`. Would be nice if this was a public method so we can correlate the response with the query.

```
responses = ms.execute()

for response in responses:
    print("Results for query %r." % response.search.query)
    for hit in response:
        print(hit.title)
```

miguelgrinberg:
Fixes #1223

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -663,7 +663,7 @@ If you need to execute multiple searches at the same time you can use the
     responses = ms.execute()
 
     for response in responses:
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM.

Issue #1533, opened 2021-09-22
add type annotations


chdsbd:
Hi!

Adding type annotations to this library would allow for better auto complete and type checking with tools like mypy and pyright.

Would you accept a PR that adds type annotations?

heinrichreimer:
This would be incredibly useful for the Elasticsearch DSL, as it would mean truly type-safe access to Elasticsearch from Python!

heinrichreimer:
Ideally, the package should then also be declared as a [PEP 561 compliant stub package](https://mypy.readthedocs.io/en/stable/installed_packages.html#installed-packages).

CosmicTiger:
Hello everyone. I would like to know if there's any continuation regarding this. I often time has this issue from the VSCode Linting and if is possible I would like to help and contribute in solving this issue if you would let me too after a brief explanation on how the project works. 
![image](https://github.com/elastic/elasticsearch-dsl-py/assets/30613501/4ea02f05-9dce-43dc-84ce-d384e0240ac0)

pquentin:
Hello! As the new maintainer, I am in favor of adding types, but do not want one huge pull request. Instead, it should be done incrementally, as described in https://sethmlarson.dev/tests-arent-enough-case-study-after-adding-types-to-urllib3. I will be happy to review small pull requests, adding types to small files first. Thanks!

miguelgrinberg:
This is going to be out with the 8.15 release.

PR #1821, opened 2024-05-09
Add Type hints to `query.py`

Caiofcas:
Starts the process of adding type hints, as proposed in #1533 .

As this is the first PR in that sense some more general things are also added here, such as the `type_check` section in `noxfile.py`. The implementation of it was based on https://sethmlarson.dev/tests-arent-enough-case-study-after-adding-types-to-urllib3 and linked issues in that blog post (specifically: https://github.com/urllib3/urllib3/pull/1924#discussion_r474283738). The filtering of mypy errors allows us to use the `--strict` flag while also implementing type hints incrementally.

One other change in `noxfile.py` is adding the `E704` code to the ignores list, as this conflicts with one line `@overload  def` statements, which is the default way black formats them (https://github.com/PyCQA/pycodestyle/issues/1036 and https://github.com/PyCQA/flake8/issues/1921). To me it seems harmless enough but don't know if this is the preferred solution.

For the actual typing, there is one change that is worth mentioning in more detail:

In `query.py::Q`, the function performed a check to see if `name_or_query` was an instance of `collections.abc.Mapping`, afterwards calling the methods `copy` and `popitem` on it. Both of these methods are not present on the `Mapping` [base class](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), with the `copy` method not present in any collection. 

I changed the call to `.copy()` to a call to `copy.deepcopy` from the standardlib, and changed the check to `collections.abc.MutableMapping` since this is the class that defines the `.popitem()` module.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.12"
+          python-version: "3.8"
       - name: Install dependencies
         run: |
           python3 -m pip install nox
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,21 @@ jobs:
       - name: Lint the code
         run: nox -s lint
 
+  type_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python3 -m pip install nox
+      - name: Lint the code
+        run: nox -s type_check
+
   docs:
     runs-on: ubuntu-latest
     steps:
--- a/elasticsearch_dsl/function.py
+++ b/elasticsearch_dsl/function.py
@@ -16,11 +16,13 @@
 #  under the License.
 
 import collections.abc
+from typing import Dict
 
 from .utils import DslBase
 
 
-def SF(name_or_sf, **params):
+# Incomplete annotation to not break query.py tests
+def SF(name_or_sf, **params) -> "ScoreFunction":
     # {"script_score": {"script": "_score"}, "filter": {}}
     if isinstance(name_or_sf, collections.abc.Mapping):
         if params:
@@ -86,7 +88,7 @@ class ScriptScore(ScoreFunction):
 class BoostFactor(ScoreFunction):
     name = "boost_factor"
 
-    def to_dict(self):
+    def to_dict(self) -> Dict[str, int]:
         d = super().to_dict()
         if "value" in d[self.name]:
             d[self.name] = d[self.name].pop("value")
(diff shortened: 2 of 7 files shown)


miguelgrinberg:
@Caiofcas Awesome, thank you so much! I will review this over the next few days.

miguelgrinberg:
Looks good overall, but I have a couple of notes/comments:

- Note that I merged main into your branch. This is to get a fix I just added for a flaky test that was causing failures in this PR.
- Would it make sense to add type hints to the query tests as well, or were you planning to do those separately once more of the library is typed?
- I suppose we also want to run the type check as part of the CI?
- This is probably for later, but I'd like to figure out a way to have better typing than `Any` for the kwargs. I will discuss this internally, it would be nice if we could somehow import these from the appropriate functions or methods in the low-level client. This would be a significant effort that should be handled separately from this, but I think once we start adding types people are not going to be happy that these are missing.

Caiofcas:
> Looks good overall, but I have a couple of notes/comments:
> 
>  * Note that I merged main into your branch. This is to get a fix I just added for a flaky test that was causing failures in this PR.

Thanks! was confused about what was the failure.
 
>  * Would it make sense to add type hints to the query tests as well, or were you planning to do those separately once more of the library is typed?

Yeah, had a look and it was not too much work to add them, so it probably makes sense to add them in this PR. The only thing is I had to add some "looser" type annotations to `DslBase.to_dict` and  `functions.SF` due to their use in the tests. Also added a flag in `mypy.ini` so the reexport of `SF` doesn't have to be commented everytime.

Don't know what the opinion of you guys is on these 'looser' but temporary annotations in exchange for a more reviewable PR, will defer to your judgement.

>   * I suppose we also want to run the type check as part of the CI? 

Added it.

>   * This is probably for later, but I'd like to figure out a way to have better typing than `Any` for the kwargs. I will discuss this internally, it would be nice if we could somehow import these from the appropriate functions or methods in the low-level client. This would be a significant effort that should be handled separately from this, but I think once we start adding types people are not going to be happy that these are missing.

Yeah, I don't like it as well but it did seem like to much to try to tackle now, since it would require more precise typing of DslBase and some of the baser classes like that. I think the Any allows people to handle it sufficiently well in their own checks.

miguelgrinberg:
@Caiofcas I tested the type checking on Python 3.8, which is the oldest we support, and this is the output:

```
❯ nox -s type_check
nox > Running session type_check
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/type_check
nox > python -m pip install mypy '.[develop]'
nox > Session type_check aborted:
elasticsearch_dsl/query.py:234: error: "list" is not subscriptable, use "typing.List" instead  [misc].
```

I think this isn't a huge deal because it is an internal part of the library. But if this was a public declaration it would be. What do you think about switching the type check task to use 3.8 so that we don't inadvertently start using newer typing constructs that will not work on the older Pythons?

Other than that I really have nothing more to suggest, so we can go ahead and merge this.

miguelgrinberg (approved):

miguelgrinberg:
Thank you so much!

PR #1822, opened 2024-05-10
Mock embedding model to address flaky vector tests

miguelgrinberg:
The recent vector tests download a model from hugging face. This PR mocks the model to prevent these random failures.


miguelgrinberg:
Merging this, very low risk.

PR #1827, opened 2024-05-17
Add Type hints to `function.py`

Caiofcas:
Continuation of #1821 (issue #1533 ).

Notable changes:

- Add `_JsonSafeTypes` in `utils.py` in an attempt to properly catch all cases in `.to_dict()` methods.
- Same `Mapping` -> `MutableMapping` and `.copy()` -> `deepcopy()` changes in `SF()` function as was done in `Q()`.
- Still having some `Any` types in kwargs and params, maybe create a new ticket to tighten these up after initial type hints are done? don't know if I can just go by the test cases and assume types not present in them are invalid, the way the functions are written it seems purposefully loose.

### DIFF
--- a/elasticsearch_dsl/function.py
+++ b/elasticsearch_dsl/function.py
@@ -17,12 +17,27 @@
 
 import collections.abc
 from copy import deepcopy
-from typing import Dict, Optional, ClassVar, Union, MutableMapping, Any
+from typing import Any, ClassVar, Dict, MutableMapping, Optional, Union, overload
 
 from .utils import DslBase, _JSONSafeTypes
 
 
-def SF(name_or_sf: Union[str, "ScoreFunction", MutableMapping[str, Any]], **params: Any) -> "ScoreFunction":
+@overload
+def SF(name_or_sf: MutableMapping[str, Any]) -> "ScoreFunction": ...
+
+
+@overload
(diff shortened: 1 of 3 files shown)


miguelgrinberg:
> Add `_JsonSafeTypes` in utils.py in an attempt to properly catch all cases in `.to_dict()` methods.

Does this type need the `_`prefix?

> Same `Mapping` -> `MutableMapping` and `.copy()` -> `deepcopy()` changes in `SF()` function as was done in `Q()`.

Looks good.

> Still having some `Any` types in kwargs and params, maybe create a new ticket to tighten these up after initial type hints are done? don't know if I can just go by the test cases and assume types not present in them are invalid, the way the functions are written it seems purposefully loose.

Yes, this will have to be treated as a separate change. You are correct, this was designed many years ago with the idea of being completely dynamic. I think in the end what I'm going to end up doing is add the `kwargs` explicitly in all of these, but let's not worry about that for this effort.

miguelgrinberg:
@Caiofcas Merging this change, but note that I have renamed `_JSONSafeTypes` to `JSONType`. There is a bunch of other internal things in utils.py (and many more in other modules) that do not use the `_` prefix, so to me it seems more consistent to not use it.

But anyway, you are making great progress! :)

miguelgrinberg (approved):

PR #1623, opened 2022-09-24
Deep paging

reese-allison:
Implements deep paging using search_after and Point in Time (PIT). This can be used to page through all results much more cheaply than using the Elasticsearch scan method. PITs are cheaper to open, so this should be safe for user requests, and can be used as a drop in replacement for scan in many cases.

Closes #1329

PR #1829, opened 2024-05-20
Added Response.search_after() method

miguelgrinberg:
The intention is to rebuild the change proposed in https://github.com/elastic/elasticsearch-dsl-py/pull/1623 as independent pieces supporting `search_after`, point-in-time, and ultimately a better version of `scan()` that uses them.

This PR implements `search_after` support in the `Response` class, also available through the `Search` class for convenience.

### DIFF
--- a/tests/test_integration/_async/test_search.py
+++ b/tests/test_integration/_async/test_search.py
@@ -126,6 +126,52 @@ async def test_scan_iterates_through_all_docs(async_data_client):
 
 
 @pytest.mark.asyncio
+async def test_search_after(async_data_client):
+    page_size = 7
+    s = AsyncSearch(index="flat-git")[:page_size].sort("authored_date")
+    commits = []
+    while True:
+        r = await s.execute()
+        commits += r.hits
+        if len(r.hits) < page_size:
+            break
+        s = r.search_after()
+
+    assert 52 == len(commits)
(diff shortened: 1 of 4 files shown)


pySilver:
Nice work! it would be nice to have search_after shortcut built-in into this lib.

tests/test_integration/_async/test_search.py
```diff
@@ -125,6 +125,52 @@ async def test_scan_iterates_through_all_docs(async_data_client):
...
+        s = r.search_after()
+
+    assert 52 == len(commits)
+    assert {d["_id"] for d in FLAT_DATA} == {c.meta.id for c in commits}
+
+
+@pytest.mark.asyncio
+async def test_search_after_no_search(async_data_client):
+    s = AsyncSearch(index="flat-git")
+    with raises(ValueError):
```
pquentin, line 147: nit: using the `match` parameter of `pytest.raises` would help clarify what errors we're actually raising here and in the following tests

pquentin (approved):
Thanks! LGTM.

tests/test_integration/_async/test_search.py
miguelgrinberg: Sure, I'll add that as well. Thanks!

PR #1835, opened 2024-06-06
Add type hints to wrappers.py

Caiofcas:
Adds type hints to `wrappers.py` and starts to add typing to `utils.AttrDict`. Continuation of #1533 

This one is more complex then previous typing MRs, so I'm open to suggestions. The main points are:

- The definition of the `SupportsComparison` TypeAlias to ensure that the operators in Range can be compared. This could posibly be avoided by utilizing the  [typeshed](https://github.com/python/typeshed) library, but at this point it seems like overkill to add it as a development dependency for only these 4/5 types.

- The use of `cast` in some instances where I needed to convey information to mypy that can be easily verified by the code validations, but it's not inferred by the type-checking. I feel that is the cleaner solution and feel that the 3 cases where it's used are justifiable.

Leave it up to you to decide on these matters due to lacking the context on how the project maintainers feel about these kind of issues.

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -18,13 +18,25 @@
 
 import collections.abc
 from copy import copy
-from typing import Any, ClassVar, Dict, List, Optional, Type, Union
+from typing import Any, ClassVar, Dict, Generic, List, Optional, Type, TypeVar, Union
 
-from typing_extensions import Self
+from typing_extensions import Self, TypeAlias
 
 from .exceptions import UnknownDslObject, ValidationException
 
-JSONType = Union[int, bool, str, float, List["JSONType"], Dict[str, "JSONType"]]
+# Usefull types
+
+JSONType: TypeAlias = Union[
+    int, bool, str, float, List["JSONType"], Dict[str, "JSONType"]
+]
+
+
+# Type variables for internals
+
+_KeyT = TypeVar("_KeyT")
+_ValT = TypeVar("_ValT")
+
+# Constants
 
 SKIP_VALUES = ("", None)
 EXPAND__TO_DOT = True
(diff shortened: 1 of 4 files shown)


elasticsearch_dsl/utils.py
```diff
@@ -110,18 +122,20 @@ def to_list(self):
...
     """
 
-    def __init__(self, d):
+    _d_: Dict[_KeyT, _ValT]
+
+    def __init__(self, d: Dict[_KeyT, _ValT]):
         # assign the inner dict manually to prevent __setattr__ from firing
         super().__setattr__("_d_", d)
 
-    def __contains__(self, key):
```
miguelgrinberg, line 124: Would this type be `_KeyT` instead of `object`?

miguelgrinberg:
> The definition of the SupportsComparison TypeAlias to ensure that the operators in Range can be compared. This could posibly be avoided by utilizing the [typeshed](https://github.com/python/typeshed) library, but at this point it seems like overkill to add it as a development dependency for only these 4/5 types.

I don't understand this. Mypy bundles typeshed, so aren't all these types already available? This seems to work:

```python
if TYPE_CHECKING:
    from _operator import _SupportsComparison
```

I'm going to go ahead and add a commit here with this simplification, since I have it working here. If you think this is missing something feel free to remove this commit and continue with your version.

miguelgrinberg:
@Caiofcas Also added some minor changes to the casts, to help mypy do a better job of it and have less explicit casts in the code. Let me know what you think of all these changes.

elasticsearch_dsl/utils.py
Caiofcas: I kept is as `object` to be consistent with the `dict.__contains__()` typing, conceptually I think is also not a type error to ask if an object of a wrong type is in the dictionary, is just gonna be false.

Caiofcas:
> I don't understand this. Mypy bundles typeshed, so aren't all these types already available? This seems to work:
> 
> ```python
> if TYPE_CHECKING:
>     from _operator import _SupportsComparison
> ```
> 
> I'm going to go ahead and add a commit here with this simplification, since I have it working here. If you think this is missing something feel free to remove this commit and continue with your version.

Hm, thanks for catching this, I did miss that ofc mypy would have it installed if it was giving the error with these types :upside_down_face: .

> @Caiofcas Also added some minor changes to the casts, to help mypy do a better job of it and have less explicit casts in > the code. Let me know what you think of all these changes.

They seem ok to me.

```
if d is None:
      data = cast(Dict[ComparisonOperators, RangeValT], kwargs)
else:
      data = d
```

If you wanna be super strict this cast is technically wrong since the validation that the keys are `ComparisonOperators` and not `str` happens in the following lines, but doing it this way does avoid the 2 casts. Maybe one day can be removed if mypys inference gets a bit stronger, but I don't think it makes a difference now, can keep it like this (I agree that it looks cleaner)

Caiofcas:
Looks good to me, thanks for the improvements!

miguelgrinberg (approved):

PR #1845, opened 2024-06-13
Use Python types to declare document fields

miguelgrinberg:
With this change, `Document` and `InnerDoc` instances can optionally use Python type hints instead or in addition to DSL field objects, and inherit dataclass-style behaviors which allow type checkers to correctly infer types.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -9,7 +9,7 @@ layer for your application.
 For more comprehensive examples have a look at the examples_ directory in the
 repository.
 
-.. _examples: https://github.com/elastic/elasticsearch-dsl-py/tree/master/examples
+.. _examples: https://github.com/elastic/elasticsearch-dsl-py/tree/main/examples
 
 .. _doc_type:
 
@@ -66,14 +66,14 @@ settings in elasticsearch (see :ref:`life-cycle` for details).
 Data types
 ~~~~~~~~~~
 
-The ``Document`` instances should be using native python types like
+The ``Document`` instances use native python types like ``str`` and
 ``datetime``. In case of ``Object`` or ``Nested`` fields an instance of the
-``InnerDoc`` subclass should be used just like in the ``add_comment`` method in
-the above example where we are creating an instance of the ``Comment`` class.
+``InnerDoc`` subclass is used, as in the ``add_comment`` method in the above
+example where we are creating an instance of the ``Comment`` class.
 
 There are some specific types that were created as part of this library to make
-working with specific field types easier, for example the ``Range`` object used
-in any of the `range fields
+working with some field types easier, for example the ``Range`` object used in
+any of the `range fields
 <https://www.elastic.co/guide/en/elasticsearch/reference/current/range.html>`_:
 
 .. code:: python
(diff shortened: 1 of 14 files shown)


miguelgrinberg:
@pquentin I suggest you start from the new documentation in persistence.rst when reviewing this.

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+Here are some simple examples:
+
+.. code:: python
+
+    from typing import Optional
+
+    class Post(Document):
+        title: str                      # same as Text(required=True)
+        created_at: Optional[datetime]  # same as Date(required=False)
+        published: bool                 # same as Boolean(required=True)
```
pquentin, line 119: Would it be helpful to call out that using type hints has different syntax than assignments somewhere?

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+   * - ``int``
+     - ``Integer(required=True)``
+   * - ``float``
+     - ``Float(required=True)``
+   * - ``bytes``
+     - ``Binary(required=True)``
+   * - ``datetime``
+     - ``Date(required=True)``
+   * - ``date``
+     - ``Date(format="yyyy-MM-dd", required=True)``
```
pquentin, line 142: This is very helpful, should we also add List and Optional?

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+    class Comment(InnerDoc):
+        content: str = Text(analyzer='snowball')
+
+The standard ``Optional`` modifier from the Python ``typing`` package can be
+used to change a typed field from required to optional. The ``List`` modifier
+can be added to a field to convert it to an array, similar to using the
+``multi=True`` argument on the field object.
+
+When using type hints as above, subclasses of ``Document`` and ``InnerDoc``
+inherit some of the behaviors associated with Python dataclasses. If
```
pquentin, line 189: I was nodding along until here, and was surprised to see that subclasses using type hints "inherited some of the behaviors associated with Python dataclasses". We don't know which ones. If it's not important which behaviors, maybe we should not mention them at all? Or maybe link to `dataclass_transform` docs?

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+necessary, the ``mapped_field()`` wrapper can be used on the right side of a
+typed field declaration, enabling dataclass options such as ``default`` or
+``default_factory`` to be included:
```
pquentin, line 192: Starting with "If necessary" feels weird because I have no idea what's necessary or not. What do you think of this wording?

```suggestion
inherit some of the behaviors associated with Python dataclasses. To add
per-field dataclass options such as ``default`` or ``default_factory`` , the
``mapped_field()`` wrapper can be used on the right side of a typed field
declaration:
```

Also, maybe we should mention that this is similar to `dataclass.field()` (with an Intersphinx link) and maybe explain why `dataclass.field()` does not work here?

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+    class MyDocument(Document):
+        title: M[str]
+        created_at: M[datetime] = mapped_field(default_factory=datetime.now)
+
+    doc = MyDocument()
+    # doc.title is typed as "str"
+    # MyDocument.title is typed as "InstrumentedField"
+
+Note that the ``M`` type hint does not provide any runtime behavior, it just
+provides additional typing declarations for type checkers.
```
pquentin, line 236: Should we encourage using it everywhere type hints are used? If not, then should it only be added to specific fields to fix typing issues? It feels some sort of short guideline would help here.

docs/persistence.rst
```diff
@@ -103,6 +103,148 @@ in any of the `range fields
...
+
+The ``InstrumentedField`` objects returned when fields are accessed as class
+attributes are proxies for the field instances that can be used anywhere a
+field needs to be referenced, such as when specifying sort options in a
+``Search`` object:
+
+.. code:: python
+
+    # sort by creation date descending, and title ascending
+    s = MyDocument.search().sort(-MyDocument.created_at, MyDocument.title)
```
pquentin, line 246: Are you saying that this sort() call would not be accepted by mypy without `M`?

elasticsearch_dsl/document_base.py
```diff
@@ -28,26 +42,174 @@ def __init__(self, *args, **kwargs):
         self.args, self.kwargs = args, kwargs
 
 
+class InstrumentedField:
+    """Proxy object for a mapped document field.
+
+    An object of this instance is returned when a field is access as a class
```
pquentin, line 48:
```suggestion
    An object of this instance is returned when a field is accessed as a class
```

elasticsearch_dsl/document_base.py
```diff
@@ -28,26 +42,174 @@ def __init__(self, *args, **kwargs):
...
+
+    def __getattr__(self, attr):
+        f = None
+        try:
+            f = self._field[attr]
+        except KeyError:
+            pass
+        if isinstance(f, Field):
+            return InstrumentedField(f"{self._name}.{attr}", f)
+        return getattr(self._field, attr)
```
pquentin, line 72: `getattr(self._field, attr)` is equivalent to `self._field[attr]`. Did you intend to specify None as a default?

Maybe we should add https://github.com/PyCQA/flake8-bugbear which would have caught it (rule B009)

elasticsearch_dsl/search_base.py
```diff
@@ -619,8 +623,16 @@ def source(self, fields=None, **kwargs):
         if fields and kwargs:
             raise ValueError("You cannot specify fields and kwargs at the same time.")
 
+        def ensure_strings(fields):
+            if isinstance(fields, list):
+                return [str(f) for f in fields]
+            elif isinstance(fields, dict):
+                return {k: ensure_strings(v) for k, v in fields.items()}
+            else:
+                return str(fields)
```
pquentin, line 632: Why didn't we need this before?

pquentin:
By the way, I forgot to mention it, but this is very cool :+1:

docs/persistence.rst
miguelgrinberg: I'll clarify this. For some reason I tried to avoid mentioning `dataclass_transform` here, but now that you make me think about this, it is best to be clear about what this means exactly.

docs/persistence.rst
miguelgrinberg: Good call, I'll add them to the table.

docs/persistence.rst
miguelgrinberg: Yes, this is the only reason we need `M`, unfortunately. Without it, the class attributes are typed the same as the instance attributes.

elasticsearch_dsl/search_base.py
miguelgrinberg: Because using class attributes in sort is a new feature I'm introducing with this change. In the current version only string representations of fields are supported and accessing a class attribute in the current version returns `AttributeError`.

I like the idea of using the class attributes to reference to fields, like SQLAlchemy does, and that means that we may get a mix of strings and the new `InstrumentedAttribute` instances whenever field names are referenced.

docs/persistence.rst
miguelgrinberg: I will clarify this. I think the guideline should be that `M` helps achieve zero errors/warnings for people who type check their code, especially if they use strict mode. And this is only for one use case, which is when you use class attributes to reference a field name. For any other usages, or if you don't care about type checkers, there is no need to complicate your code with it.

docs/persistence.rst
miguelgrinberg: There is an error at runtime if you mistakenly use a DSL type as a type hint. If you do this:

```python
class MyDoc(Document):
    title: Text
```

You get this error:

```
TypeError: Cannot map type <class 'elasticsearch_dsl.field.Text'>
```

And if you do this:

```python
class MyDoc(Document):
    title: Text(required=True)
```

The error is:

```
TypeError: Cannot map type Text()
```

elasticsearch_dsl/document_base.py
miguelgrinberg: You are right, I already have the attribute in `f`, so here I just need to return it.

miguelgrinberg:
@pquentin Thanks for all the feedback! I think I have addressed all of your notes. I rewrote the `__getattr__` implementation in `InstrumentedField` in a way that is much more clear (to me, at least), and I have also added a unit test that explicitly tries all the possible ways this class can be used. Let me know what you think.

docs/persistence.rst
pquentin: Did you decide against adding `Optional`?

docs/persistence.rst
```diff
@@ -157,38 +166,40 @@ that class. When the ``InnerDoc`` subclass is wrapped with ``List``, a
...
+        category: str = Keyword(required=True)
 
 This form can also be used when additional options need to be given to
 initialize the field, such as when using custom analyzer settings:
 
 .. code:: python
 
     class Comment(InnerDoc):
-        content: str = Text(analyzer='snowball')
+        content: str = Text(analyzer='snowball', required=True)
```
pquentin, line 190: Do we still need this part given the first example already has an option (`required=True`)?

pquentin (commented):
Thanks! Only left two questions.

docs/persistence.rst
miguelgrinberg: I didn't. Sorry, I added List, but left out `Optional`. I'm adding it.

docs/persistence.rst
miguelgrinberg: Okay, yes, I see that this makes less sense now. I'll convert the previous example to optional, so that this is the first time additional options are added.

docs/persistence.rst
miguelgrinberg: Okay, so I tried a couple of different formats and found that growing the table makes it too long and more difficult to visually parse. So in the end I kept the table small with just the required variants of the Python native types, and right after the table I show examples of `Optional`, `List`, `Object` and `Nested`. Let me know what you think.

pquentin (approved):
Thanks! LGTM, even if the table doesn't have List or Optional. Maybe better to explain it afterwards as you did indeed.

honzakral:
fwiw it would imho be much better to have `str` default to `Keyword`. It is more common in database use cases and`Text`, without the ability to specify other options like `fields` or `analyzer` is of very limited usefulness and almost never used.

PR #1849, opened 2024-06-24
Added type hints to aggs.py, analysis.py, connections.py and field.py

miguelgrinberg:
Continuing on the work that @Caiofcas started, this PR adds type hints to a few more source files, their tests, and anything else that needed to be typed in other files.


elasticsearch_dsl/utils.py
```diff
@@ -233,7 +245,7 @@ def __init__(cls, name, bases, attrs):
         # skip for DslBase
         if not hasattr(cls, "_type_shortcut"):
             return
-        if cls.name is None:
```
miguelgrinberg, line 236: This is (I think) the only logic change I've made. The `name` attribute defaults to `None` in base classes, and is set to the name of the entity in each subclass. To avoid having to type this as `Optional[str]` and then having to deal with null checks everywhere, I change the default to an empty string.

elasticsearch_dsl/aggs.py
```diff
@@ -70,48 +90,58 @@ def A(name_or_agg, filter=None, **params):
 class Agg(DslBase):
     _type_name = "agg"
     _type_shortcut = staticmethod(A)
-    name = None
+    name = ""
```
miguelgrinberg, line 93: See note about this change in utils.py below.

elasticsearch_dsl/aggs.py
```diff
@@ -70,48 +90,58 @@ def A(name_or_agg, filter=None, **params):
...
+    def result(self, search: "SearchBase", data: Any) -> AttrDict[str, Any]:
         return AggResponse(self, search, data)
 
 
 class AggBase:
-    _param_defs = {
+    if TYPE_CHECKING:
+        aggs: Dict[str, Agg]
+        _base: Agg
+    _params: Dict[str, Any]
```
pquentin, line 114: Why isn't `params` under `if TYPE_CHECKING:`? (By the way, I don't usually bother with such guards unless it avoids an import at runtime or fixes a cyclic dependency)

elasticsearch_dsl/mapping_base.py
```diff
@@ -57,7 +58,7 @@ def field(self, name, *args, **kwargs):
         self.properties[name] = construct_field(*args, **kwargs)
         return self
 
-    def _collect_fields(self):
+    def _collect_fields(self) -> Generator[Field, None, None]:
```
pquentin, line 61: Should this be an `Iterator[Field]` then?

mypy.ini
```diff
@@ -1,6 +1,2 @@
 [mypy]
```
pquentin, line 1: Why did you need to pass the options explicitly in noxfile.py? This will break direct usage of `mypy` in the future.

Now, I suppose there is a valid reason, which brings me to my next question: why keep this file at all?

pquentin (commented):
OK, so that's indeed a lot of `type: ignore`, `Any` and `cast`. But it gets the job done, and we can always improve later.

mypy.ini
miguelgrinberg: Oh, this was just a matter of preference, and to be consistent with other tools that have their options in the noxfile. I will remove this file.

elasticsearch_dsl/mapping_base.py
miguelgrinberg: Okay, sure, that works too.

elasticsearch_dsl/aggs.py
miguelgrinberg: You are right, this is inconsistent and in this case there is no point in having a guard.

pquentin (approved):
Thanks! LGTM.

@Caiofcas Feel free to add review comments after the fact, too.

PR #1851, opened 2024-06-26
type hints for document, search_base and utils classes

miguelgrinberg:
This is another somewhat large PR that tackles the difficult area of typing search results. I started adding types to document_base.py and {_async|_sync}/document.py, plus any other related files that were necessary, which are many.

I set out to support fully typed results when you are searching using a document class. Here is an example from the vectors.py example, where the search automagically returns `WorkplaceDoc` instances as search results, just because the search object was created by calling `WorkplaceDoc.search()`:

Screenshot 2024-06-27 at 4 33 08 PM

This is implemented using generics. The search class is now represented as `Search[DocumentClass]`, where results are returned as instances of `DocumentClass`. If using `Search` alone, the default `Hit` object is used for typing, matching the runtime behavior.

Adding this required plumbing the generic type through a large part of the library. I also made some other changes along the way, to help me reach the goal of fully typing the document source files:

- I simplified the `AttrDict[K, V]` generic class to just `AttrDict[V]`. Using any key types other than `str` creates lots of issues with the attribute dunder methods, which require attribute names to be strings. Really this class is never used with key types other than `str` anyway.
- I've added a `AsyncUsingType`, `UsingType` and `AnyUsingType` to represent valid values to pass to the `using` argument that many classes have. These can be connection names or client instances, for the async and sync variants. The 3rd option is used in generic places where async vs. sync does not matter.
- The `AttrList` class in utils.py does not have a generic type yet, and maybe it should, but I'm going to leave that for later.
- I was using vectors.py as test bed to verify that typed searches were working fine, so I also added all the types to that example.
- There are many things that still need to be typed in responses, for now I have only concentrated on typing the top level search results. Things such as aggregations, inner hits, etc. need to be done yet and are all `Any` at this time, I'll eventually get to do them as well.


elasticsearch_dsl/_async/document.py
```diff
@@ -69,12 +80,23 @@ class AsyncDocument(DocumentBase, metaclass=AsyncIndexMeta):
     Model-like class for persisting documents in elasticsearch.
     """
 
+    if TYPE_CHECKING:
+        _index: AsyncIndex
+
     @classmethod
-    def _get_connection(cls, using=None):
+    def _get_using(cls, using: Optional[AsyncUsingType] = None) -> AsyncUsingType:
```
miguelgrinberg, line 87: _get_using() was in the `DocumentBase` class, but now that it needs to be typed it needs to come down to the async/sync specialized classes.

elasticsearch_dsl/_async/document.py
```diff
@@ -113,7 +143,13 @@ async def get(cls, id, using=None, index=None, **kwargs):
         return cls.from_es(doc)
 
     @classmethod
-    async def exists(cls, id, using=None, index=None, **kwargs):
+    async def exists(
+        cls,
+        id: str,
+        using: Optional[AsyncUsingType] = None,
+        index: Optional[str] = None,
+        **kwargs: Any,
```
miguelgrinberg, line 151: FYI I am not worrying about kwargs at this time, leaving most of them as `Any` to be reviewed later.

elasticsearch_dsl/_async/document.py
```diff
@@ -317,9 +369,13 @@ async def update(
             doc_meta["if_seq_no"] = self.meta["seq_no"]
             doc_meta["if_primary_term"] = self.meta["primary_term"]
 
+        i = self._get_index(index)
+        assert i is not None
```
miguelgrinberg, line 373: This little dance is just to help mypy understand that the index is never going to be `None`.

elasticsearch_dsl/_async/document.py
```diff
@@ -126,12 +162,18 @@ async def exists(cls, id, using=None, index=None, **kwargs):
         ``Elasticsearch.exists`` unchanged.
         """
         es = cls._get_connection(using)
-        return await es.exists(index=cls._default_index(index), id=id, **kwargs)
+        return bool(await es.exists(index=cls._default_index(index), id=id, **kwargs))
```
miguelgrinberg, line 165: Suppressing the low level client response here and returning a bool, which seems more clear to me.

elasticsearch_dsl/_async/search.py
```diff
@@ -16,28 +16,32 @@
...
 from elasticsearch.helpers import async_scan
+from typing_extensions import TypeVar
 
 from ..async_connections import get_connection
-from ..response import Response
+from ..response import Hit, Response
 from ..search_base import MultiSearchBase, SearchBase
 from ..utils import AttrDict
 
+_R = TypeVar("_R", default=Hit)
```
miguelgrinberg, line 30: This `_R` typevar is inserted into pretty much every search or response related class. I've used the version of `TypeVar` from typing_extensions because it allows you to use a default.

elasticsearch_dsl/document_base.py
```diff
@@ -58,11 +64,14 @@ class MyDocument(Document):
...
-    def __init__(self, name, field):
+    def __init__(self, name: str, field: Field):
         self._name = name
         self._field = field
 
-    def __getattr__(self, attr):
+    # note that the return value type here assumes classes will only be used to
+    # access fields (I haven't found a way to make this type dynamic based on a
+    # decision taken at runtime)
+    def __getattr__(self, attr: str) -> "InstrumentedField":
```
miguelgrinberg, line 74: I'm unhappy about this one, but I don't think it can be handled properly when only at runtime we'll know which type we are returning.

elasticsearch_dsl/document_base.py
```diff
@@ -372,7 +377,7 @@ def __repr__(self):
             ),
         )
 
-    def to_dict(self, include_meta=False, skip_empty=True):
+    def to_dict(self, include_meta: bool = False, skip_empty: bool = True) -> Dict[str, JSONType]:  # type: ignore[override]
```
miguelgrinberg, line 380: I have a bunch of methods that change the return type in subclasses, and mypy does not like them. I'm ignoring these overrides one by one for now, because it seems useful to have mypy detect overrides that may be unintended.

elasticsearch_dsl/response/__init__.py
```diff
@@ -15,61 +15,89 @@
...
-    def success(self):
+    def success(self) -> bool:
         return self._shards.total == self._shards.successful and not self.timed_out
 
     @property
-    def hits(self):
+    def hits(self) -> List[_R]:
         if not hasattr(self, "_hits"):
-            h = self._d_["hits"]
+            h = cast(AttrDict[JSONType], self._d_["hits"])
```
miguelgrinberg, line 97: This is an common issue in many places that I think comes from a not great `Dict[str, JSONType]` typing that we use for data coming from the API. I'm not sure if there is a better way to type this so that it does not require casting each time an item is extracted, at some point I will think about that.

elasticsearch_dsl/response/__init__.py
```diff
@@ -123,33 +153,42 @@ def search_after(self):
         """
         if len(self.hits) == 0:
             raise ValueError("Cannot use search_after when there are no search results")
-        if not hasattr(self.hits[-1].meta, "sort"):
+        if not hasattr(self.hits[-1].meta, "sort"):  # type: ignore
```
miguelgrinberg, line 156: This and the next one below will probably start working once I type more of the response object.

elasticsearch_dsl/faceted_search_base.py
```diff
@@ -316,36 +337,44 @@ def search(self):
...
-    def __getitem__(self, k):
+    def __getitem__(self, k: Union[int, slice]) -> "FacetedSearchBase[_R]":
         self._s = self._s[k]
         return self
 
-    def __iter__(self):
-        return iter(self._s)
+    # TODO this goes in async/sync subclasses!
+    # def __iter__(self) -> Iterator[Response[_R]]:
+    #    return iter(self._s)
```
pquentin, line 373: Did you intend to leave this TODO?

elasticsearch_dsl/search_base.py
```diff
@@ -632,24 +722,27 @@ def ensure_strings(fields):
...
 
-        for key, value in kwargs.items():
-            if value is None:
-                try:
-                    del s._source[key]
-                except KeyError:
-                    pass
-            else:
-                s._source[key] = ensure_strings(value)
+        if isinstance(s._source, dict):
```
pquentin, line 731: Should kwargs be typed as some kind of mapping to avoid the runtime isinstance call? `kwargs` is always going to be a dict.

elasticsearch_dsl/utils.py
```diff
@@ -19,33 +19,47 @@
...
-# Usefull types
+if TYPE_CHECKING:
+    from elastic_transport import ObjectApiResponse
+    from elasticsearch import AsyncElasticsearch, Elasticsearch
+
+    from .document_base import DocumentOptions
+    from .field import Field
+    from .index_base import IndexBase
+
+# Useful types
```
pquentin, line 51: nit: Thanks for fixing the typo, but that does not seem like a very useful comment!

elasticsearch_dsl/utils.py
```diff
@@ -496,51 +525,52 @@ def __get_field(cls, name):
                     return cls._index._mapping[name]
                 except KeyError:
                     pass
+        return None
```
pquentin, line 528: Should this only apply to `except KeyError`? The try clause already returns. (Since this is only needed to help developers reason about the code, after all)

```suggestion
          return None
```

examples/vectors.py
```diff
@@ -97,9 +98,9 @@ def get_embedding(cls, input: str) -> List[float]:
             cls._model = SentenceTransformer(MODEL_NAME)
         return cast(List[float], list(cls._model.encode(input)))
 
-    def clean(self):
+    def clean(self) -> None:
         # split the content into sentences
-        passages = nltk.sent_tokenize(self.content)
+        passages = cast(List[str], nltk.sent_tokenize(self.content))
```
pquentin, line 103: Offtopic, but I would consider in the future other sentence tokenizers like the Unicode one or spaCy (e.g. the NLTK tokenizer fails on `e.g.` :) )

tests/test_validation.py
```diff
@@ -37,7 +37,10 @@ class Author(InnerDoc):
     email = Text(required=True)
 
     def clean(self):
-        print(self, type(self), self.name)
+        if not self.name:
+            raise ValidationException("name is missing")
+        if not self.email:
+            raise ValidationException("email is missing")
         if self.name.lower() not in self.email:
```
pquentin, line 44: I like making it clear when branches are exclusive:

```suggestion
        if not self.email:
            raise ValidationException("email is missing")
        elif self.name.lower() not in self.email:
```

(Also applies to the next if)

pquentin (changes requested):
Thanks! This looks great.

elasticsearch_dsl/faceted_search_base.py
miguelgrinberg: No, that is code that needs to be deleted. Thanks!

elasticsearch_dsl/_async/document.py
```diff
@@ -186,14 +229,19 @@ async def mget(
             error_ids = [doc["_id"] for doc in error_docs]
             message = "Required routing not provided for documents %s."
             message %= ", ".join(error_ids)
-            raise RequestError(400, message, error_docs)
+            raise RequestError(400, message, error_docs)  # type: ignore
```
miguelgrinberg, line 232: @pquentin Forgot to mention this one. I had to type ignore these because the first arg of `RequestError` (which is just an alias for the transport's `APIError`) is typed as a `str`, and I believe not intended to be the status code.

```python
>>> r = RequestError(400, "bad request", {"foo":"bar"})
>>> r.status_code
  File "/Users/miguelgrinberg/Documents/dev/elasticsearch-dsl-py/venv/lib/python3.12/site-packages/elasticsearch/exceptions.py", line 46, in status_code
    return self.meta.status
           ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'status'
```

Sounds like this is a bug that we should fix at some point.

elasticsearch_dsl/search_base.py
pquentin: OK, that's annoying. Thanks.

elasticsearch_dsl/_async/document.py
pquentin: Yes, this is a bug. ApiError is actually quite annoying to raise manually now: https://github.com/elastic/elastic-transport-python/blob/39488817cd5da824101322e40652d17938f0acac/tests/test_exceptions.py#L36-L41

PR #1853, opened 2024-07-01
type hints for remaining elasticsearch_dsl classes

miguelgrinberg:
This change adds type hints to the remaining files in the library (not including tests and examples, which I'll do separately).

Interesting things to note:

- The `exists_type()`, `upgrade()` and `get_upgrade()` methods of the index do not exist in the 8.x low-level client, so I have removed them from this client as well.
- I have modified the type check action in noxfile.py to only use an include list for tests and examples, which currently have partial support for types. For the main library all files are all considered typed without the need to add them to the list.
- It is quite likely that some types will have to be refined once the tests and examples are typed.
- I have to think about adding some `Protocol` definitions to the response classes, so that they can have better types and need less/no casts. This will come later, so for now I have continued making the necessary casts to silence mypy.

### DIFF
--- a/elasticsearch_dsl/_async/index.py
+++ b/elasticsearch_dsl/_async/index.py
@@ -168,14 +168,14 @@ class AsyncIndex(IndexBase):
 
     async def create(
         self, using: Optional[AsyncUsingType] = None, **kwargs: Any
-    ) -> None:
+    ) -> "ObjectApiResponse[Any]":
         """
         Creates the index in elasticsearch.
 
         Any additional keyword arguments will be passed to
         ``Elasticsearch.indices.create`` unchanged.
         """
-        await self._get_connection(using).indices.create(
+        return await self._get_connection(using).indices.create(
             index=self._name, body=self.to_dict(), **kwargs
         )
 
@@ -185,7 +185,9 @@ class AsyncIndex(IndexBase):
         )
         return bool(state["metadata"]["indices"][self._name]["state"] == "close")
 
-    async def save(self, using: Optional[AsyncUsingType] = None) -> None:
+    async def save(
+        self, using: Optional[AsyncUsingType] = None
+    ) -> "Optional[ObjectApiResponse[Any]]":
         """
         Sync the index definition with elasticsearch, creating the index if it
         doesn't exist and updating its settings and mappings if it does.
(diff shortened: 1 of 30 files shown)


elasticsearch_dsl/_async/index.py
```diff
@@ -141,24 +166,26 @@ def updateByQuery(self, using=None):
             index=self._name,
         )
 
-    async def create(self, using=None, **kwargs):
+    async def create(
+        self, using: Optional[AsyncUsingType] = None, **kwargs: Any
+    ) -> None:
```
pquentin, line 171: Why `-> None`? The response could be useful I guess: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#create-index-wait-for-active-shards

elasticsearch_dsl/_async/search.py
```diff
@@ -158,16 +164,20 @@ async def iterate(self, keep_alive="1m"):
                     yield hit
                 if len(r.hits) == 0:
                     break
-                s = r.search_after()
+                s = s.search_after()
```
pquentin, line 167: This does not change the behavior, right?

elasticsearch_dsl/faceted_search_base.py
```diff
@@ -361,10 +364,7 @@ def __init__(
 
         self._s = self.build_search()
 
-    def count(self) -> int:
-        return self._s.count()
```
pquentin, line 365: Why remove `count()`? This sounds like it could be useful somehow

pquentin (approved):
Thanks! LGTM. None of my comments are blocking.

elasticsearch_dsl/_async/search.py
miguelgrinberg: No, the type hint in the search call is more specific than the one in the response, so it works better for type checking. But they do the same thing.

elasticsearch_dsl/faceted_search_base.py
miguelgrinberg: Because it is in the wrong place. This is something I missed when we did the async changes, `count()` should be a coroutine in the async case, so it has to move down to the async/sync subclasses of `SearchBase`. Look in `_async/faceted_search.py` and `_sync/faceted_search.py` for it.

elasticsearch_dsl/_async/index.py
miguelgrinberg: Yes, you are right, here I can return the response object from the low-level client.

pquentin (approved):

PR #1858, opened 2024-07-04
make fields typed as lists optional

miguelgrinberg:
I'm trying to type the examples, and another common issue I'm seeing is related to lists, either those created with `multi=True` or by using a `Nested` field.

The issue is with the state of the `required` option of the field. When using the non-typed field creation option this isn't a problem. Both `Text(multi=True)` and `Nested(SomeDoc)` work well with their default setting of `required=False`.

When using the new type hint definitions, however, the default for a typed field is to be required, and to make a field not required you have to wrap the type with `Optional`. This is okay for a scalar field, but for a list it creates confusion, because in terms of Python typing it is expected that an empty list is a valid value for a required field, but a required DSL field considers an empty list the same as `None` and does not allow it.

With this change, I'm changing the field creation logic to make sure that any fields that are created via type hints have `required=False` even when `Optional` is not given in the type hint, so that they work in the expected way and allow the empty list as a valid value.

No behavior changes are made for fields that are created without type hints.

### DIFF
--- a/elasticsearch_dsl/document_base.py
+++ b/elasticsearch_dsl/document_base.py
@@ -166,24 +166,11 @@ class DocumentOptions:
         field_defaults = {}
         for name in fields:
             value = None
-            if name in attrs:
-                # this field has a right-side value, which can be field
-                # instance on its own or wrapped with mapped_field()
-                value = attrs[name]
-                if isinstance(value, dict):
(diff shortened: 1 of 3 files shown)


miguelgrinberg:
Closing. I have included this change in #1859

pquentin (approved):
Thanks! LGTM.

PR #1859, opened 2024-07-04
Type hints for tests and examples

miguelgrinberg:
Type hints for all remaining tests and examples.

There are some notes embedded in the code, but the main changes are:

- All the code, tests and examples now pass mypy strict 🎉 
- I've added pyright to the type checking pass, but only for examples, to make sure vscode/vim/etc. do not produce any errors. This also passes.
- I'm closing #1858 as those changes are included in this PR as well. The description of the change in that PR is still useful for you to read.
- In test files that work with `Document` instances I have silenced a number of mypy errors that come as a result of using the old way of creating document fields without type hints. This is because I want to make sure these continue to be supported and tested.
- For consistency, I have added a generic type to `AttrList`. In all current usages we are doing `AttrList[Any]` and `AttrDict[Any]`, but this might change in the future.

Work that remains wrt types:

- Responses are typed with `Any`. It would be useful to create specific types for all the fields.
- There are inconsistencies in typing lists with `List` vs. `Sequence` and dicts with `Dict` and `Mapping` or `MutableMapping` that I will try to clean up at some point.

### DIFF
--- a/elasticsearch_dsl/__init__.py
+++ b/elasticsearch_dsl/__init__.py
@@ -16,7 +16,7 @@
 #  under the License.
 
 from . import connections
-from .aggs import A
+from .aggs import A, Agg
 from .analysis import analyzer, char_filter, normalizer, token_filter, tokenizer
 from .document import AsyncDocument, Document
 from .document_base import InnerDoc, M, MetaField, mapped_field
@@ -81,7 +81,8 @@ from .field import (
 from .function import SF
 from .index import AsyncIndex, AsyncIndexTemplate, Index, IndexTemplate
 from .mapping import AsyncMapping, Mapping
-from .query import Q
+from .query import Q, Query
+from .response import AggResponse, Response, UpdateByQueryResponse
 from .search import (
     AsyncEmptySearch,
     AsyncMultiSearch,
@@ -99,6 +100,8 @@ __version__ = VERSION
 __versionstr__ = ".".join(map(str, VERSION))
 __all__ = [
     "A",
+    "Agg",
+    "AggResponse",
     "AsyncDocument",
     "AsyncEmptySearch",
     "AsyncFacetedSearch",
@@ -158,11 +161,13 @@ __all__ = [
     "Object",
     "Percolator",
     "Q",
+    "Query",
     "Range",
     "RangeFacet",
     "RangeField",
     "RankFeature",
     "RankFeatures",
+    "Response",
     "SF",
     "ScaledFloat",
     "Search",
@@ -174,6 +179,7 @@ __all__ = [
     "TokenCount",
     "UnknownDslObject",
     "UpdateByQuery",
+    "UpdateByQueryResponse",
     "ValidationException",
     "analyzer",
     "char_filter",
--- a/elasticsearch_dsl/_async/search.py
+++ b/elasticsearch_dsl/_async/search.py
@@ -16,7 +16,16 @@
 #  under the License.
 
 import contextlib
-from typing import Any, AsyncIterator, Dict, Iterator, List, Optional, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AsyncIterator,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    cast,
+)
 
 from elasticsearch.exceptions import ApiError
 from elasticsearch.helpers import async_scan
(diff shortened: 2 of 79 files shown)


elasticsearch_dsl/__init__.py
```diff
@@ -99,6 +100,8 @@
 __versionstr__ = ".".join(map(str, VERSION))
 __all__ = [
     "A",
+    "Agg",
```
miguelgrinberg, line 103: There are some classes that were not directly exported at the package level because the user rarely needs to directly create them. Now that we have to write type hints these are sometimes needed, so I decided to add them so that people don't go out to fish them out from their internal locations.

elasticsearch_dsl/_async/search.py
```diff
@@ -175,6 +185,10 @@ class AsyncMultiSearch(MultiSearchBase[_R]):
 
     _using: AsyncUsingType
 
+    if TYPE_CHECKING:
+
+        def add(self, search: AsyncSearch[_R]) -> Self: ...  # type: ignore[override]
```
miguelgrinberg, line 190: There are a few methods that override their definition in the parent class. I have type ignored them for now, because at some point I want to revisit them and see if they can be changed to match.

elasticsearch_dsl/aggs.py
```diff
@@ -140,7 +140,12 @@ def __iter__(self) -> Iterable[str]:
         return iter(self.aggs)
 
     def _agg(
-        self, bucket: bool, name: str, agg_type: str, *args: Any, **params: Any
+        self,
+        bucket: bool,
+        name: str,
+        agg_type: Union[Dict[str, Any], Agg[_R], str],
```
miguelgrinberg, line 146: mypy was perfectly happy with my naive `agg_type: str` type, but pyright flagged an issue in one of the examples that pointed at this as the problem.

elasticsearch_dsl/utils.py
```diff
@@ -111,10 +111,10 @@ def __ne__(self, other: Any) -> bool:
     def __getitem__(self, k: Union[int, slice]) -> Any:
         l = self._l_[k]
         if isinstance(k, slice):
-            return AttrList(l, obj_wrapper=self._obj_wrapper)
+            return AttrList[_ValT](l, obj_wrapper=self._obj_wrapper)  # type: ignore[arg-type]
```
miguelgrinberg, line 114: This is extremely annoying from mypy. Here we know that `l` is of type `List[_ValT]`, but mypy thinks it can also be a single element of type `_ValT` because it does not see the check for the slice.

examples/alias_migration.py
```diff
@@ -38,24 +38,28 @@
...
 PATTERN = ALIAS + "-*"
 
 
 class BlogPost(Document):
-    title = Text()
-    published = Date()
-    tags = Keyword(multi=True)
-    content = Text()
+    if TYPE_CHECKING:
+        _id: int
```
miguelgrinberg, line 51: This is a trick to get the dataclass support to accept these extra arguments in the constructor.

tests/_async/test_document.py
```diff
@@ -15,12 +15,18 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
+# this file creates several documents using bad or no types because
+# these are still supported and should be kept functional in spite
+# of not having appropriate type hints. For that reason the comment
+# below disables many mypy checks that fails as a result of this.
+# mypy: disable-error-code="assignment, index, arg-type, call-arg, operator, comparison-overlap, attr-defined"
```
miguelgrinberg, line 22: This file has a lot of errors because of documents that are defined using old syntax withtout the type hints. I felt it was best to silence those errors and keep some of those older syntax examples.

examples/alias_migration.py
pquentin: Since this is an example, should we explain the trick?

noxfile.py
```diff
@@ -38,8 +36,6 @@
     "tests/test_query.py",
     "tests/test_utils.py",
     "tests/test_wrappers.py",
-    "examples/vectors.py",
-    "examples/async/vectors.py",
 )
```
pquentin, line 39: Should TYPED_FILES be removed?

examples/alias_migration.py
miguelgrinberg: Yes, I should add a comment here, thanks.

noxfile.py
miguelgrinberg: Yes, I forgot.

PR #1862, opened 2024-07-09
Add index creation to aggregations example

miguelgrinberg:
The aggregations example (composite_agg.py) assumes an index called `git` has been created and populated in advance, which in practice makes it impossible for most people to get the example to run.

The git index is one of the datasets used by the integration tests. In this change I have added the logic to create and populate the index, so that the example can be standalone like all the other examples.

### DIFF
--- a/tests/test_integration/test_data.py
+++ b/tests/test_integration/test_data.py
@@ -25,9 +25,6 @@ user_mapping = {
 
 FLAT_GIT_INDEX: Dict[str, Any] = {
     "settings": {
-        # just one shard, no replicas for testing
-        "number_of_shards": 1,
-        "number_of_replicas": 0,
         # custom analyzer for analyzing file paths
         "analysis": {
             "analyzer": {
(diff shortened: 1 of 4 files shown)


tests/test_integration/test_data.py
```diff
@@ -19,99 +19,91 @@
...
             },
+        }
+    },
+}
+
+GIT_INDEX: Dict[str, Any] = {
+    "settings": {
+        # just one shard, no replicas for testing
+        "number_of_shards": 1,
+        "number_of_replicas": 0,
```
pquentin, line 63: So I know this was already here, but this seems pointless. After all examples are here to share good practices. And I would be surprised if this resulted in significant savings.

pquentin (approved):

tests/test_integration/test_data.py
miguelgrinberg: Yes, I agree, I'll remove it.

pquentin (approved):

PR #1864, opened 2024-07-10
Add support for bulk document operations

miguelgrinberg:
This change adds `Document.bulk` and `AsyncDocument.bulk`. These work similarly to the bulk helper in the low-level client, but the provided generator must return document instances, and the document's `clean()` method is called before adding to the bulk request.

### DIFF
--- a/elasticsearch_dsl/_async/document.py
+++ b/elasticsearch_dsl/_async/document.py
@@ -16,9 +16,20 @@
 #  under the License.
 
 import collections.abc
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AsyncIterable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
 
 from elasticsearch.exceptions import NotFoundError, RequestError
+from elasticsearch.helpers import async_bulk
 from typing_extensions import Self, dataclass_transform
 
 from .._async.index import AsyncIndex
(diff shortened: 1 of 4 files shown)


elasticsearch_dsl/_async/document.py
```diff
@@ -438,3 +449,73 @@ async def save(
...
+
+        :arg actions: a generator that returns document instances to be indexed,
+            bulk operation dictionaries.
+        :arg using: connection alias to use, defaults to ``'default'``
+        :arg index: elasticsearch index to use, if the ``Document`` is
+            associated with an index this can be omitted.
+        :arg validate: set to ``False`` to skip validating the documents
+        :arg skip_empty: if set to ``False`` will cause empty values (``None``,
+            ``[]``, ``{}``) to be left on the document. Those values will be
+            stripped out otherwise as they make no difference in elasticsearch.
```
pquentin, line 474:
```suggestion
            stripped out otherwise as they make no difference in Elasticsearch.
```

elasticsearch_dsl/_async/document.py
```diff
@@ -438,3 +449,73 @@ async def save(
...
+        skip_empty: bool = True,
+        **kwargs: Any,
+    ) -> Tuple[int, Union[int, List[Any]]]:
+        """
+        Allows to perform multiple indexing operations in a single request.
+
+        :arg actions: a generator that returns document instances to be indexed,
+            bulk operation dictionaries.
+        :arg using: connection alias to use, defaults to ``'default'``
+        :arg index: elasticsearch index to use, if the ``Document`` is
```
pquentin, line 469:
```suggestion
        :arg index: Elasticsearch index to use, if the ``Document`` is
```

pquentin (approved):
Thanks! LGTM.

PR #1865, opened 2024-07-20
export async_connections at the package level

miguelgrinberg:
The `async_connections` module was missing as a top-level package export.


pquentin (approved):
Thanks! LGTM.

PR #1866, opened 2024-07-20
Instrumented fields in queries

miguelgrinberg:
Instances of `InstrumentedField` might pop up in queries and aggregations, for example:

```python
s = s.query('knn', field=QuoteDoc.embedding, query_vector=model.encode(q).tolist())
s.aggs.bucket('tags', 'terms', field=QuoteDoc.tags, size=100)
```

With this change they are handled appropriately.


pquentin (approved):
Thanks! LGTM.

PR #1867, opened 2024-07-29
Add py typed file

miguelgrinberg:
None of the typing commits we've done included the py.typed file, so here it is.


pquentin (approved):
Thanks! LGTM.

PR #1877, opened 2024-08-13
Removed `__orig_class__` attribute out of AttrDict instances

### LINKED ISSUE #1876: v8.15.0 breaks things related to type hints?
Hello,

some of my existing code breaks when upgrading from v8.14.0 to v8.15.0 as follows:

```
self = elasticsearch_dsl.utils.AttrDict[typing.Any], attr = 'doc_count'

    def __getattr__(self, attr):
        if attr in {'__name__', '__qualname__'}:
            return self._name or self.__origin__.__name__

        # We are careful for copy and pickle.
        # Also for simplicity we don't relay any dunder names
        if '__origin__' in self.__dict__ and not _is_dunder(attr):
>           return getattr(self.__origin__, attr)
E           AttributeError: type object 'AttrDict' has no attribute 'doc_count'

/usr/lib/python3.12/typing.py:1185: AttributeError
```

Greetings,

jgb

miguelgrinberg:
Fixes #1876 

This change addressed a regression introduced in 8.15.0 as part of the refactor to support type hints.

The problem is in `AttrDict`, which can get confused and inject an attribute named `__orig_class__` in the underlying dict. For example:

```python
>>> from typing import Any
>>> from elasticsearch_dsl import AttrDict
>>> d = AttrDict[Any]({})
>>> d.to_dict()
{'__orig_class__': elasticsearch_dsl.utils.AttrDict[typing.Any]}
```

This happens because Python adds the `__orig_class__` attribute to all instances that were created with explicit generic arguments given. The logic in `AttrDict.__setattr__` to detect if an attribute is part of the dictionary or a real attribute of the instance does not work for `__orig_class__`.

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -215,7 +215,11 @@ class AttrDict(Generic[_ValT]):
         del self._d_[key]
 
     def __setattr__(self, name: str, value: _ValT) -> None:
-        if name in self._d_ or not hasattr(self.__class__, name):
+        # the __orig__class__ attribute has to be treated as an exception, as
+        # is it added to an object when it is instantiated with type arguments
+        if (
+            name in self._d_ or not hasattr(self.__class__, name)
+        ) and name != "__orig_class__":
             self._d_[name] = value
         else:
             # there is an attribute on the class (could be property, ..) - don't add it as field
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks! LGTM. I see #1878 fixes the CI.

PR #1881, opened 2024-08-15
Added support for the `semantic_text` field and `semantic` query type

miguelgrinberg:
This change adds support for semantic text, introduced in Elasticsearch 8.15, plus an example application. Unfortunately running this application requires a somewhat beefy ES instance, so I'm not going to add integration tests.

### DIFF
--- a/examples/async/semantic_text.py
+++ b/examples/async/semantic_text.py
@@ -1,0 +1,150 @@
+#  Licensed to Elasticsearch B.V. under one or more contributor
+#  license agreements. See the NOTICE file distributed with
+#  this work for additional information regarding copyright
+#  ownership. Elasticsearch B.V. licenses this file to you under
+#  the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


examples/async/semantic_text.py
```diff
@@ -0,0 +1,150 @@
...
+            category=data["category"],
+        )
+        await doc.save()
+
+    # refresh the index
+    await WorkplaceDoc._index.refresh()
+
+
+async def search(query: str) -> dsl.AsyncSearch[WorkplaceDoc]:
+    return WorkplaceDoc.search()[:5].query(
```
pquentin, line 111: nit: Doing `[:5]` before the query call is a bit confusing IMO.

examples/async/semantic_text.py
```diff
@@ -0,0 +1,150 @@
...
+        "semantic",
```
pquentin, line 112: Any reason not to use the class here? One advantage is that it avoids typos.

pquentin (approved):
Thanks! LGTM.

examples/async/semantic_text.py
miguelgrinberg: No reason at all, except that historically it has been preferred to use the names. I actually prefer the classes myself and started using them on my tests. I'll update this.

PR #1890, opened 2024-08-28
Autogenerate Query classes

miguelgrinberg:
This change adds a code-generation phase to this client. In this first phase, code generation is applied to two files:
- query.py: All the sub-classes of `Query` are auto-generated and include docstrings, type hints and a constructor with explicit arguments. All classes have `kwargs` that are backwards compatible.
- interfaces.py: this is a new file in which all the types that are referenced by attributes of the `Query` sub-classes are defined.

Things that are not currently implemented but will likely be added later:
- Shortcut properties.
- Properties defined with generics in the specification are given an `Any` type for now.
- All analyzers are given a `Union[str, Dict[str, Any]]` type for now.
- Aggregations, Responses, possibly other things will be added in follow-up PRs.


elasticsearch_dsl/faceted_search_base.py
```diff
@@ -173,27 +173,26 @@ def __init__(
...
     def get_value_filter(self, filter_value: FilterValueType) -> Query:
         f, t = self._ranges[filter_value]
-        limits = {}
+        limits: Dict[str, Any] = {}
         if f is not None:
             limits["gte"] = f
         if t is not None:
             limits["lt"] = t
 
-        return Range(_expand__to_dot=False, **{self._params["field"]: limits})
```
miguelgrinberg, line 182: The `Range` query class now has a more reasonable format that is typing friendly, which uses `Range(field, value, **additional_kwargs)`. The old form works, but makes mypy unhappy.

elasticsearch_dsl/function.py
```diff
@@ -17,9 +17,18 @@
...
+    ClassVar,
+    Dict,
+    Literal,
+    MutableMapping,
+    Optional,
+    Union,
+    overload,
+)
 
-from .utils import DslBase
```
miguelgrinberg, line 22: `NotSet` is a class I use to make arguments are not set. `NOT_SET` is the only instance of class `NotSet`. They are used extensively in generated code.

elasticsearch_dsl/function.py
```diff
@@ -137,3 +146,29 @@ class Gauss(ScoreFunction):
 
 class Exp(ScoreFunction):
     name = "exp"
+
+
+class DecayFunction(AttrDict[Any]):
```
miguelgrinberg, line 151: It seemed overkill to autogenerate just a single function here, so I coded this one by hand using the autogenerated style.

elasticsearch_dsl/utils.py
```diff
@@ -151,6 +158,7 @@ class AttrDict(Generic[_ValT]):
     """
 
     _d_: Dict[str, _ValT]
+    RESERVED: Dict[str, str] = {"from_": "from"}
```
miguelgrinberg, line 161: AttrDict now automatically recognizes `from` keys and exposes them as `from_`.

noxfile.py
```diff
@@ -57,8 +57,9 @@ def test(session):
 
 @nox.session(python="3.12")
 def format(session):
-    session.install("black~=24.0", "isort", "unasync", "setuptools")
+    session.install("black~=24.0", "isort", "unasync", "setuptools", ".[develop]")
     session.run("python", "utils/run-unasync.py")
+    session.run("python", "utils/generator.py", env={"PYTHONPATH": "./"})
```
miguelgrinberg, line 62: Code generation runs in the `format` command.

tests/_async/test_search.py
```diff
@@ -479,22 +479,26 @@ def test_complex_example() -> None:
 def test_reverse() -> None:
     d = {
         "query": {
-            "filtered": {
```
miguelgrinberg, line 482: The `filtered` query was deprecated around ES 2.x and does not exist anymore in the schema. I replaced it with `bool`.

tests/test_query.py
```diff
@@ -124,7 +124,7 @@ def test_query_clone() -> None:
 
 
 def test_bool_converts_its_init_args_to_queries() -> None:
-    q = query.Bool(must=[{"match": {"f": "value"}}])
+    q = query.Bool(must=[{"match": {"f": "value"}}])  # type: ignore
```
miguelgrinberg, line 127: I could have used the class here, but the whole point of the test is to make sure the dict is equivalent to the class version, so I ignored the typing errors.

utils/templates/interfaces.py.tpl
```diff
@@ -0,0 +1,45 @@
+from typing import Any, Dict, List, Literal, Mapping, Union
+from elasticsearch_dsl.document_base import InstrumentedField
+from elasticsearch_dsl import function as f, interfaces as i, Query
+from elasticsearch_dsl.utils import AttrDict, NotSet, NOT_SET
+
+PipeSeparatedFlags = str
```
miguelgrinberg, line 6: For now I'm not worrying about `PipeSeparatedFlags` and consider it a string. But we could come up with a more Pythonic way to generate a list of flags.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,361 @@
...
+        self,
+        {% if k.kwargs and not k.has_field and not k.has_fields %}
+        *,
+        {% endif %}
+        {% for kwarg in k.kwargs %}
+        {{ kwarg.name }}: {{ kwarg.type }} = NOT_SET,
+        {% endfor %}
+        **kwargs: Any
+    ):
+        {% if k.name == "FunctionScore" %}
```
miguelgrinberg, line 183: The original `FunctionScore` function had some custom handling for the `functions` attribute, which is a shortcut property. Since it is the only class that attempted to do this, I'm manually keeping that logic here. In the future we may consider moving this logic to the code generator.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,361 @@
...
+        if isinstance(functions, NotSet):
+            functions = []
+            for name in ScoreFunction._classes:
+                if name in kwargs:
+                    functions.append({name: kwargs.pop(name)})  # type: ignore
+        {% elif k.has_field %}
```
miguelgrinberg, line 189: A class that is marked as `has_field` is one that has a single key dictionary with the key being a field. This exists in several classes. For this type of structure the constructor takes two positional arguments, the field and the value.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,361 @@
...
+        if not isinstance(_field, NotSet):
+            kwargs[str(_field)] = _value
+        {% elif k.has_fields %}
```
miguelgrinberg, line 192: The `has_fields` flag is for classes that are dictionaries with fields as keys. The constructor takes a single positional argument that is a dictionary.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,352 @@
+#  Licensed to Elasticsearch B.V. under one or more contributor
```
miguelgrinberg, line 1: This template is somewhat complex, because it has a farly significant manually generated code at the top, which are the generic parts of the file up to and including the `Query` base class.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,352 @@
...
+        return Bool(should=[self, other])
+
+    def __and__(self, other: "Query") -> "Query":
+        # make sure we give queries that know how to combine themselves
+        # preference
+        if hasattr(other, "__rand__"):
+            return other.__rand__(self)
+        return Bool(must=[self, other])
+
+
```
miguelgrinberg, line 144: The dynamic portion of the template starts here.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,352 @@
...
+        {% endif %}
+        super().__init__(
+            {% if not k.is_single_field and not k.is_multi_field %}
+            {% for arg in k.args %}
+            {{ arg.name }}={{ arg.name }},
+            {% endfor %}
+            {% endif %}
+            **kwargs
+        )
+
```
miguelgrinberg, line 205: There are a handful of Query sub-classes that had extra helper methods to make them more Pythonic. I'm just adding those here to keep things compatible with current code.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,352 @@
...
+        {% endfor %}
+        **kwargs: Any
+    ):
+        {% if k.name == "FunctionScore" %}
+        if isinstance(functions, NotSet):
+            functions = []
+            for name in ScoreFunction._classes:
+                if name in kwargs:
+                    functions.append({name: kwargs.pop(name)})  # type: ignore
+        {% elif k.is_single_field %}
```
miguelgrinberg, line 189: Special handling for single-key dicts.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,352 @@
...
+        if not isinstance(_field, NotSet):
+            kwargs[str(_field)] = _value
+        {% elif k.is_multi_field %}
```
miguelgrinberg, line 192: Special handling for multi-key dicts.

pquentin (commented):
This is a first review after spending one hour scratching my head on this pull request. 

While the inline comments help a lot, there are many things I don't understand yet. We're going to need the review of someone that maintains a typed client as well at some point. @JoshMock and @Anaethelion come to mind, as they have types in their client and are familiar with Python.

elasticsearch_dsl/utils.py
```diff
@@ -97,6 +97,13 @@ def _recursive_to_dict(value: Any) -> Any:
         return value
 
 
+class NotSet:
+    pass
+
+
+NOT_SET = NotSet()
+
+
```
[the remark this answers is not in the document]
miguelgrinberg: I should have checked the main client to see how this was solved there. I'll adapt to the existing solution.

elasticsearch_dsl/function.py
```diff
@@ -137,3 +146,29 @@ class Gauss(ScoreFunction):
...
+        offset: Any = NOT_SET,
+        scale: Any = NOT_SET,
+        origin: Any = NOT_SET,
+        multi_value_mode: Union[
+            Literal["min", "max", "avg", "sum"], "NotSet"
+        ] = NOT_SET,
+        **kwargs: Any,
+    ):
+        if not isinstance(decay, NotSet):
+            kwargs["decay"] = decay
```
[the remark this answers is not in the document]
miguelgrinberg: Mainly to make mypy happy. I think the `DEFAULT` solution in the main client and the transport is better than my approach, so I think I'll be able to just compare with `!= DEFAULT` without confusing mypy.

utils/generator.py
```diff
@@ -0,0 +1,483 @@
...
+    print(f"Generated {filename}.")
+
+
+def generate_interfaces_py(schema, interfaces, filename):
+    """Generate interfaces.py"""
+    classes = {}
+    for interface in interfaces:
+        if interface == "PipeSeparatedFlags":
+            continue  # handled as a special case
+        schema.reset_interfaces()
```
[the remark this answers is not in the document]
miguelgrinberg: This was an attempt to get a concise list of newly discovered interfaces. By resetting the list I would just get new ones. But I agree that this isn't a great solution. I already have a list of interfaces so I should be able to append the new ones without having to do this.

elasticsearch_dsl/utils.py
miguelgrinberg: How are you thinking that would work? It seemed to me the AttrDict needs to know that when I say `whateter.from_` I'm really saying `whatever["from"]`. This is going to happen in the application, not in generated code, so I have no control except from inside the AttrDict. If you can think of a better option let me know.

elasticsearch_dsl/function.py
miguelgrinberg: Replaced with the transport's `DEFAULT`.

elasticsearch_dsl/utils.py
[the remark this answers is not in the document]
miguelgrinberg: I have now adapted the generator to use the transport's `DEFAULT`, which is friendlier to mypy than my class based solution.

utils/generator.py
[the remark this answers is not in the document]
miguelgrinberg: This is now removed. There is now a single list of discovered interfaces that is constantly updated. Much cleaner solution.

utils/generator.py
```diff
@@ -0,0 +1,483 @@
...
+        elif schema_type["kind"] == "interface":
+            if schema_type["name"]["namespace"] == "_types.query_dsl":
+                # handle specific DSL classes explicitly to map to existing
+                # Python DSL classes
+                if schema_type["name"]["name"].endswith("RangeQuery"):
+                    return '"wrappers.Range[Any]"', None
+                elif schema_type["name"]["name"].endswith("ScoreFunction"):
+                    name = schema_type["name"]["name"][:-8]
+                    if name == "FieldValueFactorScore":
+                        name = "FieldValueFactor"  # Python DSL uses different name
```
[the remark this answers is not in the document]
miguelgrinberg: Yes, that is actually better than having a special case so I'm adding both names in function.py.

miguelgrinberg:
@pquentin thanks for all the feedback. I have addressed all your points I think, and made changes for each in a separate commit so that you can inspect them separately.

In addition to your feedback, I also added inherited attributes to all classes, which as I mentioned yesterday I forgot to include before. This actually made evident another thing I needed to handle, the `AdditionalProperty` behavior, which is used in a few of the query classes so that is now included as well.

Finally, I've been thinking that the name interfaces.py is odd for Python. Would renaming it to types.py be better? And also, do we want to dump all the queries and types in the root level of the package so that they can be imported directly via `from elasticsearch_dsl import X`? I recall you indicated you would prefer this way of working with these classes so let me know what you think.

utils/generator.py
```diff
@@ -0,0 +1,528 @@
...
+        classes[k["name"]] = k
+
+    classes_list = []
+    for n in classes:
+        k = classes[n]
+        try:
+            classes_list.index(k)
+            continue
+        except ValueError:
+            pass
```
pquentin, line 507: Any reason not to use [LBYL](https://docs.python.org/3/glossary.html#term-LBYL) here?

```suggestion
        if k in classes_list:
            continue
```

pquentin (commented):
Here's another round of review. I'm slowly getting to understand this better. Apologies for the reviewing delays, but this is quite daunting!

> Finally, I've been thinking that the name interfaces.py is odd for Python. Would renaming it to types.py be better?

Yes, `types` is better than `interfaces`.

> And also, do we want to dump all the queries and types in the root level of the package so that they can be imported directly via from elasticsearch_dsl import X? I recall you indicated you would prefer this way of working with these classes so let me know what you think.

Yes, I think that would fit the existing design better to dump all classes at the top-level. This will address one issue I have with the split between interfaces/types and queries: I still believe it should be an implementation detail that users don't have to care about.

elasticsearch_dsl/function.py
```diff
@@ -152,23 +154,23 @@ class DecayFunction(AttrDict[Any]):
...
-        if not isinstance(offset, NotSet):
+        if offset != DEFAULT:
             kwargs["offset"] = offset
-        if not isinstance(scale, NotSet):
-            kwargs["offset"] = scale
-        if not isinstance(origin, NotSet):
-            kwargs["offset"] = origin
-        if not isinstance(multi_value_mode, NotSet):
-            kwargs["offset"] = multi_value_mode
+        if scale != DEFAULT:
```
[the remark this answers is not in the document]
miguelgrinberg: Sure, I really have no preference, I can change it.

utils/generator.py
```diff
@@ -0,0 +1,528 @@
...
+# keys and values are in given in "{namespace}:{name}" format
+TYPE_REPLACEMENTS = {
+    "_types.query_dsl:DistanceFeatureQuery": "_types.query_dsl:DistanceFeatureQueryBase",
+}
+
+
+def wrapped_doc(text, width=70, initial_indent="", subsequent_indent=""):
+    """Formats a docstring as a list of lines of up to the request width."""
+    return textwrap.wrap(
+        (text or "No documentation available.").replace("\n", " "),
```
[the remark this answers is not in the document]
miguelgrinberg: Yes, I'll just leave it empty since this isn't adding any value.

elasticsearch_dsl/interfaces.py
```diff
@@ -0,0 +1,3283 @@
...
+    left: Union[float, "DefaultType"]
+    right: Union[float, "DefaultType"]
+
+    def __init__(
+        self,
+        *,
+        top: Union[float, "DefaultType"] = DEFAULT,
+        bottom: Union[float, "DefaultType"] = DEFAULT,
+        left: Union[float, "DefaultType"] = DEFAULT,
+        right: Union[float, "DefaultType"] = DEFAULT,
```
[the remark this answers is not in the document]
miguelgrinberg: The reason I'm putting defaults on all attributes including required ones is that I don't have complete trust that the schema is right, so I'm allowing some flexibility to be tolerant of different schema versions. For example, we are using the 8.15 schema now, but if the user connects to an older stack then what's required or optional might be different and I don't want code to break because of schema differences. The idea is to enhance the IDE experience, not to enforce arguments. Does that make sense?

elasticsearch_dsl/interfaces.py
```diff
@@ -0,0 +1,3283 @@
...
+            kwargs["_name"] = _name
+        super().__init__(**kwargs)
+
+
+class MatchPhrasePrefixQuery(QueryBase):
+    """
+    :arg analyzer: Analyzer used to convert text in the query value into
+        tokens.
+    :arg max_expansions: Maximum number of terms to which the last
+        provided term of the query value will expand.
```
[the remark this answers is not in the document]
miguelgrinberg: I'm not sure I like the idea, because server defaults can be tied to individual versions of the stack, and that will make the library do strange things if you connect to a version that is not the one that was used to generate this code.

elasticsearch_dsl/interfaces.py
```diff
@@ -0,0 +1,3283 @@
...
+
+from typing import Any, Dict, Literal, Mapping, Sequence, Union
+
+from elastic_transport.client_utils import DEFAULT, DefaultType
+
+from elasticsearch_dsl import Query
+from elasticsearch_dsl import function as f
+from elasticsearch_dsl import interfaces as i
+from elasticsearch_dsl.document_base import InstrumentedField
+from elasticsearch_dsl.utils import AttrDict
```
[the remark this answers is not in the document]
miguelgrinberg: The quotes could be removed on this file, you are correct. I'll look into that.

utils/generator.py
miguelgrinberg: Yes, the logic in this function was more complex initially and I needed to know the index, but now I've got a much simpler solution so a simple check works. Thanks.

elasticsearch_dsl/function.py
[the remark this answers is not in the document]
pquentin: Yes, please, if only for consistency with the existing generator, considering we might merge them one day.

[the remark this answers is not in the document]
miguelgrinberg: Updated.

elasticsearch_dsl/function.py
```diff
@@ -28,7 +28,9 @@
     overload,
 )
 
-from .utils import NOT_SET, AttrDict, DslBase, NotSet
+from elastic_transport.client_utils import DEFAULT, DefaultType
```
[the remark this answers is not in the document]
miguelgrinberg: I'm adding it as a dependency, seems right to do it.

elasticsearch_dsl/interfaces.py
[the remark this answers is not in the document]
pquentin: I suppose it does, but the IDE is going to tell you when something is mandatory, right? Also, yes, the specification currently has many gaps, but it's usually missing attributes, not having too much attributes. (Except for serverless, where we may have less things.) Oh, but that's why you have **kwargs.

It was tempting to say "we generate code as precise as possible, and we allow to fall-back to using a body parameter". But maybe that does not make sense here.

elasticsearch_dsl/interfaces.py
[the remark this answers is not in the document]
pquentin: I was talking about documentation only, not actually setting the default (it's a server default, not a client default).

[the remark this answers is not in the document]
miguelgrinberg: Ah, sorry! I can add the server default to the docstring, certainly.

elasticsearch_dsl/interfaces.py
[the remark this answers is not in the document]
miguelgrinberg: I'm open to discuss this, but what I decided to do is to leave the default value, but add the required flag to the docstring. For example:

```
    """
    :arg query: (required) Text, number, boolean value or date you wish to
        find in the provided field.
    """
```

I also intended to move the required fields to the top of the argument list, but now I see that this is not working correctly, so I'll fix that.

elasticsearch_dsl/interfaces.py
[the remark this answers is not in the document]
miguelgrinberg: Defaults added!

elasticsearch_dsl/interfaces.py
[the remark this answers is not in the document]
pquentin: I'm happy with the current situation, thanks. Let's not add more complexity for this!

utils/templates/query.py.tpl
pquentin: While this code is in both `query.py.tpl` and `types.py.tpl`, I'm not seeing any code in `query.py` or `types.py` using it, is that expected?

utils/templates/types.py.tpl
```diff
@@ -0,0 +1,84 @@
...
+
+from elasticsearch_dsl.document_base import InstrumentedField
+from elasticsearch_dsl import function, Query
+from elasticsearch_dsl.utils import AttrDict
+
+PipeSeparatedFlags = str
+
+
+{% for k in classes %}
+class {{ k.name }}({% if k.parent %}{{ k.parent }}{% else %}AttrDict[Any]{% endif %}):
```
pquentin, line 30: I believe using an [inline if-expression](https://jinja.palletsprojects.com/en/3.1.x/templates/#if-expression) would be clearer:

```suggestion
class {{ k.name }}({ k.parent if k.parent else "AttrDict[Any]" }}):
```

utils/generator.py
```diff
@@ -0,0 +1,544 @@
...
+            "required": arg["required"],
+        }
+        if param is not None:
+            param = {"name": arg["name"], "param": param}
+        if arg["required"]:
+            # insert in the right place so that all required arguments
+            # appear at the top of the argument list
+            i = 0
+            for i in range(len(k["args"]) + 1):
+                if i == len(k["args"]):
```
pquentin, line 293: Is this equivalent to using enumerate?

```suggestion
            i = 0
            for i, arg in enumerate(k["args"]):
```

utils/generator.py
```diff
@@ -0,0 +1,544 @@
...
+                if "parent" not in k:
+                    k["parent"] = type_["inherits"]["type"]["name"]
+                if type_["inherits"]["type"]["name"] not in interfaces:
+                    interfaces.append(type_["inherits"]["type"]["name"])
+                type_ = schema.find_type(
+                    type_["inherits"]["type"]["name"],
+                    type_["inherits"]["type"]["namespace"],
+                )
+            else:
+                break
```
pquentin, line 491: nit: I find placing the break up-front make the code easier to read.

```suggestion
        while True:
            if "inherits" not in type_ or "type" not in type_["inherits"]:
                break
                
            for arg in type_["properties"]:
                schema.add_attribute(k, arg, for_types_py=True)
            if "parent" not in k:
                k["parent"] = type_["inherits"]["type"]["name"]
            if type_["inherits"]["type"]["name"] not in interfaces:
                interfaces.append(type_["inherits"]["type"]["name"])
            type_ = schema.find_type(
                type_["inherits"]["type"]["name"],
                type_["inherits"]["type"]["namespace"],
            )
```

pquentin (approved):
Thanks! LGTM.

utils/templates/query.py.tpl
miguelgrinberg: The limitation is that it does not currently support shortcut properties. Once shortcut properties are implemented this code will be generated. So I see it as temporary, just to keep things from breaking.

utils/templates/types.py.tpl
miguelgrinberg: I didn't know you could use inline if statements in Jinja. Thanks.

utils/templates/query.py.tpl
```diff
@@ -0,0 +1,365 @@
...
+
+
+{% for k in classes %}
+class {{ k.name }}({{ parent }}):
+    """
+    {% for line in k.docstring %}
+    {{ line }}
+    {% endfor %}
+    {% if k.args %}
+    {% if k.docstring %}
```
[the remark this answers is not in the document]
miguelgrinberg: Sure, I'll see how it looks with indentation on the Jinja sections.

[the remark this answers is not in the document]
miguelgrinberg: Added the indentation.

pquentin (approved):

PR #1892, opened 2024-08-30
Added recursive to_dict support to AttrDict

### LINKED ISSUE #1520: to_dict() doesn't convert inner_hits
Maybe related to https://github.com/elastic/elasticsearch-dsl-py/issues/291, I'm attempting to dump my elasticsearch-dsl response by calling `to_dict()`, but it doesn't appear to walk the entire structure. `inner_hits` appear to be `Response` instances, but calling `to_dict()` on the top-level response doesn't convert them. Why does it convert hits but not inner_hits? Is there a simple way to convert the _entire_ response without having to pick apart the structure and manually call `to_dict()` as needed?

I'm using Python 3.8 and elasticsearch-dsl 7.4.0.

Thanks so much for this library! ❤️ It's incredibly useful.

miguelgrinberg:
Fixes #1520

The `AttrDict.to_dict()` method now has a `recursive` flag which defaults to False. If True, all contents of the dict are converted.

### DIFF
--- a/elasticsearch_dsl/utils.py
+++ b/elasticsearch_dsl/utils.py
@@ -86,6 +86,17 @@ def _wrap(val: Any, obj_wrapper: Optional[Callable[[Any], Any]] = None) -> Any:
     return val
 
 
+def _recursive_to_dict(value: Any) -> Any:
+    if hasattr(value, "to_dict"):
+        return value.to_dict()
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks! LGTM.

PR #1895, opened 2024-09-03
Allow tuples and other iterables in source() method

### LINKED ISSUE #1893: Regression: source() does not accept iterables
Prior to the following change, I could pass a tuple to `source()` calls, e.g. `source(includes=("a","b","c"))`

https://github.com/elastic/elasticsearch-dsl-py/pull/1845/files#diff-706d55be3a1b925b14bb3122ccecf00daeb04113df622f4ce445bec7d7753f5cR627

Seems like passing tuples, or any other iterable other than lists will now fail due to `isinstance(fields, list)` check.

Notice that this was before the typing update, which established the parameter type as list. This change at least needs to be documented if we are not considering supporting other collection types.

miguelgrinberg:
Fixes #1893

### DIFF
--- a/elasticsearch_dsl/search_base.py
+++ b/elasticsearch_dsl/search_base.py
@@ -37,13 +37,13 @@ from typing import (
 from typing_extensions import Self, TypeVar
 
 from .aggs import A, Agg, AggBase
+from .document_base import InstrumentedField
 from .exceptions import IllegalOperation
 from .query import Bool, Q, Query
 from .response import Hit, Response
 from .utils import _R, AnyUsingType, AttrDict, DslBase, recursive_to_dict
 
 if TYPE_CHECKING:
-    from .document_base import InstrumentedField
     from .field import Field, Object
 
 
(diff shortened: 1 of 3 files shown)


elasticsearch_dsl/search_base.py
```diff
@@ -714,10 +714,10 @@ def ensure_strings(
                 Dict[str, List[Union[str, "InstrumentedField"]]],
             ]
         ) -> Union[str, List[str], Dict[str, List[str]]]:
-            if isinstance(fields, list):
-                return [str(f) for f in fields]
-            elif isinstance(fields, dict):
+            if isinstance(fields, dict):
                 return {k: ensure_strings(v) for k, v in fields.items()}
+            elif not isinstance(fields, (str, InstrumentedField)):
+                return [str(f) for f in fields]
```
pquentin, line 720: Have you considered a more explicit check, like explicitly allowing lists and tuples or looking for for `collections.abc.Iterable`?

pquentin (approved):

elasticsearch_dsl/search_base.py
miguelgrinberg: I'll add a comment for now. Basically the way I'm thinking about this is that in this case the typing change broke how the library works at runtime. My goal was to address the runtime problem. As I mentioned back when we did the typing changes, I think we could replace specific types such as List and Dict with more generic ones in a lot of places, but when I look at that I'll do it for the whole library.

PR #1887, opened 2024-08-22
Accept Terms value to be any iterable

roxanebellot:
Fixes https://github.com/elastic/elasticsearch-dsl-py/issues/1641

CLA is signed, PR is rebased !

PR #1904, opened 2024-09-09
Accepts boost param on Terms

jairhenrique:
https://github.com/elastic/elasticsearch-dsl-py/pull/1887 introduced a bug when we tried to pass a boost in the search for terms.

This PR ignores the value list transformation when param name is `boost`.


miguelgrinberg:
Thanks! Unfortunately this fix only looks for `boost`, but for robustness we want to make sure anything that is not iterable is passed as is. I'm working on a more complete fix that will be released soon.

PR #1907, opened 2024-09-11
Fixed regression introduced in `Terms` query class

### LINKED ISSUE #1905: Terms query errors when passing the boost parameter
When I upgraded to the latest version of the Elasticsearch-DSL library, I found a bug that you cannot pass the boost parameter anymore, because it tries to parse everything as a list. This is an example that fails:

```
query.Terms(title=["Something"], boost=2)
```

Hope the problem is clear.

miguelgrinberg:
Regression introduced by #1887 
Fixes #1905

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -551,7 +551,10 @@ class Terms(Query):
     name = "terms"
 
     def _setattr(self, name: str, value: Any) -> None:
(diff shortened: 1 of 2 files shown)


jairhenrique (approved):

pquentin (approved):
Thanks! LGTM.

PR #1910, opened 2024-09-11
Fixed the the type hint for `Document.get()`

### LINKED ISSUE #1908: Document.get() documentation is misleading - not found does not return None
The [documentation for Document.get()](https://elasticsearch-dsl.readthedocs.io/en/latest/api.html#elasticsearch_dsl.Document.get) says the return type is `Optional[Self]`, and [the code](https://github.com/elastic/elasticsearch-dsl-py/blob/main/elasticsearch_dsl/_sync/document.py#L125) reads like the assumption is a document not being found can be handled without catching exceptions:

```python
        if not doc.get("found", False):
            return None
```

However, when a document is not found, rather than return None, an `elasticsearch.NotFoundError` exception is raised, necessitating the use of an exception handler in my code:

```python
        try:
            return FooDoc.get(id=params.id, index='foo-index')
        except NotFoundError:
            return None
```

It would be nice if `Document.get()` did return None if not found, but if that's not possible, the documentation should reflect the actual behavior.

miguelgrinberg:
Fixes #1908

### DIFF
--- a/elasticsearch_dsl/_async/document.py
+++ b/elasticsearch_dsl/_async/document.py
@@ -134,7 +134,7 @@ class AsyncDocument(DocumentBase, metaclass=AsyncIndexMeta):
         using: Optional[AsyncUsingType] = None,
         index: Optional[str] = None,
         **kwargs: Any,
-    ) -> Optional[Self]:
+    ) -> Self:
         """
         Retrieve a single document from elasticsearch using its ``id``.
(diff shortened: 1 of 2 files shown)


miguelgrinberg:
Closing as this is actually incorrect. The client can be configured to ignore the 404 error, and in that case the get() method should return `None`.

PR #1918, opened 2024-09-13
Auto-generate aggregation classes

miguelgrinberg:


elasticsearch_dsl/query.py
```diff
@@ -624,7 +624,9 @@ def __init__(
             Literal["multiply", "replace", "sum", "avg", "max", "min"], "DefaultType"
         ] = DEFAULT,
         functions: Union[
-            Sequence["types.FunctionScoreContainer"], Dict[str, Any], "DefaultType"
+            Sequence["types.FunctionScoreContainer"],
+            Sequence[Dict[str, Any]],
```
miguelgrinberg, line 628: This is a bug that I discovered in the generator for anything with the form `Sequence[types.Something]`, which should be given `Sequence[Dict[str, Any]]` as an alternative. And yes, maybe I should switch `Dict` to `Mapping` as well, and will try if that works at some point.

elasticsearch_dsl/search_base.py
```diff
@@ -949,7 +949,7 @@ def to_dict(self, count: bool = False, **kwargs: Any) -> Dict[str, Any]:
         d = {}
 
         if self.query:
-            d["query"] = self.query.to_dict()
+            d["query"] = recursive_to_dict(self.query)
```
miguelgrinberg, line 952: The idea with these new recursive conversions to dict is that we want to get rid of all the internal classes that we can now use for attributes and sub-attributes.

elasticsearch_dsl/utils.py
```diff
@@ -365,6 +365,7 @@ def __setattr__(self, name: str, value: Any) -> None:
 
     def _setattr(self, name: str, value: Any) -> None:
         # if this attribute has special type assigned to it...
+        name = AttrDict.RESERVED.get(name, name)
```
miguelgrinberg, line 368: Another bug that I noticed. The `AttrDict` handles `from_` and `from` correctly, but the `DslBase` class overrides this behavior and also needs to understand reserved names.

examples/async/composite_agg.py
```diff
@@ -17,19 +17,19 @@
...
 
-from elasticsearch_dsl import A, Agg, AsyncSearch, Response, async_connections
+from elasticsearch_dsl import Agg, AsyncSearch, Response, aggs, async_connections
 from tests.test_integration.test_data import DATA, GIT_INDEX
 
 
 async def scan_aggs(
     search: AsyncSearch,
-    source_aggs: Union[Dict[str, Agg], List[Dict[str, Agg]]],
+    source_aggs: Sequence[Mapping[str, Agg]],
```
miguelgrinberg, line 30: The `sources` attribute is declared as a list in the spec, so I'm adapting this example to follow that to avoid typing errors.

tests/test_aggs.py
```diff
@@ -220,7 +220,13 @@ def test_filters_correctly_identifies_the_hash() -> None:
 
 
 def test_bucket_sort_agg() -> None:
-    bucket_sort_agg = aggs.BucketSort(sort=[{"total_sales": {"order": "desc"}}], size=3)
+    bucket_sort_agg = aggs.BucketSort(sort=[{"total_sales": {"order": "desc"}}], size=3)  # type: ignore
```
miguelgrinberg, line 223: For several examples in this file I kept the original dict based solution with ignored typing errors, and right below I've added a correctly typed version, so that we make sure we are backwards compatible.

utils/generator.py
```diff
@@ -258,7 +309,9 @@ def add_attribute(self, k, arg, for_types_py=False):
             type_ = "Any"
             param = None
         if type_ != "Any":
-            if "types." in type_:
+            if 'Sequence["types.' in type_:
```
miguelgrinberg, line 312: This is not a great way to do this. I'm thinking it may be nice to have a library that understands Python type hints and is able to parse them and render them, but for now this finds all the sequences.

elasticsearch_dsl/aggs.py
```diff
@@ -222,314 +1201,2502 @@ def to_dict(self) -> Dict[str, Any]:
...
+            **kwargs,
+        )
 
-class BucketSort(Pipeline[_R]):
-    name = "bucket_sort"
+    def result(self, search: "SearchBase[_R]", data: Any) -> AttrDict[Any]:
+        return FieldBucketData(self, search, data)
+
+
+class RandomSampler(Bucket[_R]):
```
miguelgrinberg, line 3701: I'm manually adding the missing `RandomSampler` for now. I think once it is added to the schema this is going to cause a failure in the type checking, so I'll know I can remove it then.

elasticsearch_dsl/types.py
```diff
@@ -876,56 +1678,198 @@ class MatchQuery(QueryBase):
...
+        fuzzy_rewrite: Union[str, DefaultType] = DEFAULT,
+        fuzzy_transpositions: Union[bool, DefaultType] = DEFAULT,
+        lenient: Union[bool, DefaultType] = DEFAULT,
+        max_expansions: Union[int, DefaultType] = DEFAULT,
+        minimum_should_match: Union[int, str, DefaultType] = DEFAULT,
+        operator: Union[Literal["and", "or"], DefaultType] = DEFAULT,
+        prefix_length: Union[int, DefaultType] = DEFAULT,
+        zero_terms_query: Union[Literal["all", "none"], DefaultType] = DEFAULT,
+        boost: Union[float, DefaultType] = DEFAULT,
+        _name: Union[str, DefaultType] = DEFAULT,
```
pquentin, line 1695: Go home GitHub, you're drunk?

elasticsearch_dsl/aggs.py
```diff
@@ -202,17 +212,986 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+
 class Filter(Bucket[_R]):
+    """
+    A single bucket aggregation that narrows the set of documents to those
+    that match a query.
+
+    :arg filter: A single bucket aggregation that narrows the set of
+        documents to those that match a query.
+    """
```
pquentin, line 1185: I love how we're keeping essentially the same code but adding documentation!

elasticsearch_dsl/aggs.py
pquentin: `RandomSampler` is in the specification now. I've just backported to 8.15 in case it helps here.

pquentin (approved):
Thanks! LGTM.

PR #1921, opened 2024-10-04
Fixed the use of dictionaries as values in Terms query

### LINKED ISSUE #1920: Terms lookup feature no longer supported due to recent release
A recent release appears to have broken the ability to perform a "terms lookup" query which is a special subset of terms query which allows you to specify the index, id, and field so you can limit terms to those found in a specific document in another index:

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html#query-dsl-terms-lookup

I used to be able to create a "terms lookup" query like so:

```
Q('terms', **{'_id': {
    'index': 'my-other-index',
    'id': 'my-document-id',
    'path': 'my-field'
}})
```

Now, however, the query generated flattens the dictionary into a list of keys, so that what actually gets built is this:

```
"terms": {
    "_id": [
        "index",
        "id",
        "path"
    ]
}
```

This seems to be an unintended side-effect of a recent release. Please advise.

miguelgrinberg:
Fixes #1920

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -2647,7 +2647,7 @@ class Terms(Query):
 
     def _setattr(self, name: str, value: Any) -> None:
         # here we convert any iterables that are not strings to lists
-        if hasattr(value, "__iter__") and not isinstance(value, (str, list)):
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks! LGTM.

PR #1929, opened 2024-10-10
Generate types for the Elasticsearch responses

miguelgrinberg:


elasticsearch_dsl/response/__init__.py
```diff
@@ -41,11 +44,47 @@
 
 
 class Response(AttrDict[Any], Generic[_R]):
+    """An Elasticsearch response.
+
+    :arg took: (required)
```
miguelgrinberg, line 49: Sadly none of the response attributes appear to have documentation.

examples/async/completion.py
```diff
@@ -103,7 +103,7 @@ async def main() -> None:
         response = await s.execute()
 
         # print out all the options we got
-        for option in response.suggest.auto_complete[0].options:
+        for option in response.suggest["auto_complete"][0].options:
```
miguelgrinberg, line 106: To be more typing friendly I am changing all examples and tests where there is an attribute that is determined by a name given in the corresponding request to use dictionary syntax, since it is not possible to add type hints for variable names. The attribute format still works, but mypy/pyright does not like it.

utils/generator.py
```diff
@@ -119,6 +120,7 @@ def __init__(self):
         # Any interfaces collected here are then rendered as Python in the
         # types.py module.
         self.interfaces = []
+        self.response_interfaces = []
```
miguelgrinberg, line 123: The main change in the generator is that it now understands if it is generating a type that is used in requests (in other words, a type that users will construct) or in responses (a type that we construct and the user will just access).

For response types we do not generate defaults and dict alternative types.

utils/generator.py
```diff
@@ -662,4 +775,7 @@ def generate_types_py(schema, filename):
     schema = ElasticsearchSchema()
     generate_query_py(schema, "elasticsearch_dsl/query.py")
     generate_aggs_py(schema, "elasticsearch_dsl/aggs.py")
+    generate_response_init_py(schema, "elasticsearch_dsl/response/__init__.py")
+    # generate_response_hit_py(schema, "elasticsearch_dsl/response/hit.py")
+    # generate_response_aggs_py(schema, "elasticsearch_dsl/response/aggs.py")
```
miguelgrinberg, line 780: This two are not done yet.

examples/async/completion.py
pquentin: This is not super intuitive - also I don't understand "it is not possible to add type hints for variable names". Do you maybe have an example?

pquentin (approved):
Thanks! LGTM. Happy to merge this, my comments are not blocking.

examples/async/completion.py
miguelgrinberg: Sorry, let's look at this specific code. The `auto_complete` attribute here is not an attribute of anything that is in the specification, it is just a name given in the request:

```python
s = s.suggest("auto_complete", text, completion={"field": "suggest"})
```

This is a common pattern in ES requests, it happens with many things, most prominent of all in aggregations, for which I'm also switched to use dictionary syntax. Does it make more sense now? How can we provide a type hint for `auto_complete` here, given that it is only known at runtime? But if we consider `suggest` to be a dictionary, then it is easier, because `auto_complete` is now a generic string key. Here is how pyright sees `suggest`:

Screenshot 2024-10-18 at 2 32 35 PM

utils/generator.py
```diff
@@ -574,23 +637,48 @@ def interface_to_python_class(self, interface, interfaces, for_types_py=True):
         }
         ```
         """
-        type_ = schema.find_type(interface)
+        type_ = self.find_type(interface, namespace)
```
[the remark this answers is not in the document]
miguelgrinberg: Ah okay. Yes, this was something that I discovered a few days ago, the use of `schema` worked due to the global but was unintended. I thought I had all of those fixed in this PR, but it looks like I have the remaining ones in the next PR.

examples/async/completion.py
pquentin: I see, thank you 👍

PR #1932, opened 2024-10-18
Autogenerate aggregation responses and hits

miguelgrinberg:


elasticsearch_dsl/utils.py
```diff
@@ -468,7 +470,15 @@ def _clone(self) -> Self:
...
 
-class HitMeta(AttrDict[Any]):
+if TYPE_CHECKING:
+    HitMetaBase = HitBaseType
+else:
+    HitMetaBase = AttrDict[Any]
+
+
+class HitMeta(HitMetaBase):
+    inner_hits: Mapping[str, Any]
```
miguelgrinberg, line 480: I haven't found a way to type inner hints appropriately without making breaking changes. I think if we have the intetion to have proper typing we'll have to evaluate a change here.

For example, consider:

```python
hit.inner_hits["passages"][0].content
```

Here it is impossible to type the `content` attribute of the first `Passage` document, since `inner_hits` is a dictionary, and each entry in this dict maps to a different document type.

We may consider changing to something like this:

```python
hit.passages.meta.inner_hits[0].content
```

I think this may be possible to type because the inner hits belong to the document instead of being in a dict. But I'm not sure I want to get into such a big change now.

examples/async/composite_agg.py
```diff
@@ -30,7 +31,7 @@ async def scan_aggs(
     source_aggs: Sequence[Mapping[str, Agg]],
     inner_aggs: Dict[str, Agg] = {},
     size: int = 10,
-) -> AsyncIterator[Response]:
```
miguelgrinberg, line 33: `Response` here was plainly wrong, not sure how this passed mypy to be honest!

examples/async/parent_child.py
```diff
@@ -161,8 +159,6 @@ async def add_answer(
         answer = Answer(
             # required make sure the answer is stored in the same shard
             _routing=self.meta.id,
-            # since we don't have explicit index, ensure same index as self
-            _index=self.meta.index,
```
miguelgrinberg, line 165: This wasn't necessary on this example, since all the documents belong to the same index. I removed it because it creates a typing problem, because `_index` is stored as an `Index` or `AsyncIndex`, but here it was given as a string.

utils/generator.py
```diff
@@ -625,6 +638,9 @@ def interface_to_python_class(
                 "required": bool,
                 "positional": bool,
             ],
+            "buckets_as_dict": "type" # optional, only present in aggregation response
+                                      # classes that have buckets that can have a list
+                                      # or dict representation
```
miguelgrinberg, line 643: To make typing easier aggregations have `buckets` typed as a list and `buckets_as_dict` typed as a dict. You can use whichever makes sense depending on how yo request your aggregations. It's a compromise solution to avoid having to apply casts on a type that is a union of list and dict buckets.

pquentin: What do you mind with "type" here?

examples/async/composite_agg.py
```diff
@@ -52,13 +53,13 @@ async def run_search(**kwargs: Any) -> Response:
         return await s.execute()
 
     response = await run_search()
-    while response.aggregations.comp.buckets:
-        for b in response.aggregations.comp.buckets:
-            yield b
-        if "after_key" in response.aggregations.comp:
-            after = response.aggregations.comp.after_key
+    while response.aggregations["comp"].buckets:
```
pquentin, line 56: Switching from `.comp` to `["comb"]` feels like a net negative, but if that helps mypy type correctly (and raise an error on a typo) then that's OK!

pquentin (approved):
Thanks! LGTM.

examples/async/composite_agg.py
miguelgrinberg: I'm not sure I agree. To me it feels like using the dict syntax makes it clear that this is a dynamic data item that is application specific. But maybe I convinced myself this is better.

utils/generator.py
miguelgrinberg: "type" here is the type that the `buckets_as_dict` property will be. This is used on the types Jinja template:

```
    @property
    def buckets_as_dict(self) -> Mapping[str, {{ k.buckets_as_dict }}]:
        return self.buckets  # type: ignore
```

PR #1936, opened 2024-11-05
Do not map ClassVar attributes

### LINKED ISSUE #1927: Specify ignored attributes in mappings
SQLAlchemy interprets the `ClassVar` type annotation to mean a non-mapped variable which the ORM can safely ignore. This is great, since it lets us populate special fields at runtime without IDEs complaining about unresolved attributes (and, if the value is read before being set, it doesn't risk causing a runtime exception).

Is there a way to emulate this with Document mappings? We could pre-populate the fields to be empty strs/lists/dicts, but that means re-building our indices every time we want to add/remove special attributes or change their types.

miguelgrinberg:
Fixes #1927


pquentin (approved):
Thanks! LGTM.

PR #1943, opened 2024-11-11
support composable templates

miguelgrinberg:
This PR adds support for composable templates as a second templating option. The two examples that use templates have been upgraded to the newer templates.

### DIFF
--- a/elasticsearch_dsl/_async/index.py
+++ b/elasticsearch_dsl/_async/index.py
@@ -73,7 +73,7 @@ class AsyncIndexTemplate:
         )
 
 
-class AsyncNewIndexTemplate:
+class AsyncComposableIndexTemplate:
     def __init__(
         self,
         name: str,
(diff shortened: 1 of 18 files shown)


elasticsearch_dsl/_async/index.py
```diff
@@ -109,6 +150,19 @@ def as_template(
...
 
+    def as_composable_template(
+        self,
+        template_name: str,
+        pattern: Optional[str] = None,
+        priority: Optional[int] = None,
+    ) -> AsyncComposableIndexTemplate:
+        # TODO: should we allow pattern to be a top-level arg?
+        # or maybe have an IndexPattern that allows for it and have
+        # Document._index be that?
```
pquentin, line 161: It seems to me that we should either do it or remove this comment. (The current code seems perfectly usable to me.)

pquentin (approved):
Thanks! LGTM.

elasticsearch_dsl/_async/index.py
miguelgrinberg: Okay, I'll remove, since these are likely from very long ago and we clearly are fine with the way this is.

PR #1959, opened 2025-01-05
Add support for the point field

### LINKED ISSUE #1958: Add support for "point" field type
`elasticsearch_dsl/field.py` is missing "point" field type. [Point field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/point.html)

Telomeraz:
Closes #1958


miguelgrinberg:
Thanks! I added `Shape` to your PR, and also removed the examples that you created, because they could create confusion in the way that you used several different formats and an `Any` type hint. I would expect most application would stick to a single format and then use an appropriate type.

PR #1960, opened 2025-01-07
Correct typing hints for the FunctionScore query

### LINKED ISSUE #1957: Cannot use FunctionScoreContainer?
While trying to upgrade a project to version 8.17, existing code:

```
from elasticsearch_dsl.function import RandomScore
from elasticsearch_dsl.query import FunctionScore

s = Search().query(
    FunctionScore(
        functions=[
            RandomScore()
        ]
    )
)

print(s.to_dict())
```
outputs:
```
{'query': {'function_score': {'functions': [{'random_score': {}}]}}}
```

But mypy complains:
```
test.py:9: error: Argument "functions" to "FunctionScore" has incompatible type "list[RandomScore]"; expected "Sequence[FunctionScoreContainer] | Sequence[dict[str, Any]] | DefaultType"  [arg-type]
```

mypy is happy with:
```
from elasticsearch_dsl.function import RandomScore
from elasticsearch_dsl.query import FunctionScore
from elasticsearch_dsl.types import FunctionScoreContainer

s2 = Search().query(
    FunctionScore(
        functions=[
            FunctionScoreContainer(
                random_score=RandomScore()
            )
        ]
    )
)
print(s2.to_dict())
```
which fails at runtime with:
```
TypeError: unhashable type: 'FunctionScoreContainer'
```

miguelgrinberg:
Fixes #1957

The fix is simple, the code generator needed to intercept references to `FunctionScoreContainer`, halt code generation at that point and direct those references to the DSL's `ScoreFunction` class which implements this functionality idiomatically.

### DIFF
--- a/elasticsearch_dsl/query.py
+++ b/elasticsearch_dsl/query.py
@@ -612,9 +612,9 @@ class FunctionScore(Query):
 
     name = "function_score"
     _param_defs = {
+        "functions": {"type": "score_function", "multi": True},
         "query": {"type": "query"},
         "filter": {"type": "query"},
-        "functions": {"type": "score_function", "multi": True},
     }
 
     def __init__(
@@ -623,11 +623,7 @@ class FunctionScore(Query):
         boost_mode: Union[
             Literal["multiply", "replace", "sum", "avg", "max", "min"], "DefaultType"
         ] = DEFAULT,
-        functions: Union[
-            Sequence["types.FunctionScoreContainer"],
-            Sequence[Dict[str, Any]],
-            "DefaultType",
-        ] = DEFAULT,
+        functions: Union[Sequence[ScoreFunction], "DefaultType"] = DEFAULT,
         max_boost: Union[float, "DefaultType"] = DEFAULT,
         min_score: Union[float, "DefaultType"] = DEFAULT,
         query: Union[Query, "DefaultType"] = DEFAULT,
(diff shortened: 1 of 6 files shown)


elasticsearch_dsl/query.py
```diff
@@ -612,9 +612,9 @@ class FunctionScore(Query):
 
     name = "function_score"
     _param_defs = {
+        "functions": {"type": "score_function", "multi": True},
```
pquentin, line 615: For my own understanding, how does this line get added, since we removed it in query.py.tpl?

pquentin (approved):
Thanks! LGTM.

elasticsearch_dsl/query.py
miguelgrinberg: This seemed a bit "magical" to me as well when I noticed it and made me smile. It is all done by logic that is already present in the generator.

In the original version before this fix I had to put this definition in the template, because the code generator did not attempt to do anything with function scores.

But with this fix the generator recognizes `FunctionScoreContainer` and returns a DSL type for it. The return value has two components, the typing hint and the DSL type:

```python
        return "ScoreFunction", {"type": "score_function"}
```

The first returned value is used in type hints. The second one (when given, since it is optional) is added to the `_param_defs` dictionary of the class by the generator itself. After this type is seen the generator will likely see an `array_of` that wraps this type, and that adds `Sequence[...]` to the type hint, and `"multi": True` to the DSL type.

The intention is that at some point all these will be generated, but there are still some that are manually injected through the templates.

PR #1968, opened 2025-02-04
remove DSL code and redirect imports to unified Python client

miguelgrinberg:
This PR removes all the code from the Elasticsearch DSL library and replaces it with redirects to the same code now located in the Python client (https://github.com/elastic/elasticsearch-py).

I'm leaving the unit tests untouched, to confirm that the import redirects work well.

Must be merged after the unified client is released. Until then, to avoid failures the dependency on the Python client was changed to install the main branch.


pquentin (approved):
Thanks! LGTM. 🎉

Issue #1395, opened 2020-07-28
delete_by_query from Search.delete()


jdratlif:
I am trying to run a delete_by_query by using the delete() method of my Search object, but it times out.

What can I do to make sure it keeps running until completion?

This is the exception I get
elasticsearch.exceptions.ConnectionTimeout: ConnectionTimeout caused by - ReadTimeoutError(HTTPSConnectionPool(host='localhost', port=9200): Read timed out. (read timeout=300)) 

Trying to add a `request_timeout` parameter led me to a 

    search.delete(request_timeout=0)
TypeError: delete() got an unexpected keyword argument 'request_timeout'

TypeError exception.

The docs for the delete_by_query call say that by default it keeps running until it is finished, so there must be something I need to do with timeout to make sure it does.

Thanks.

sethmlarson:
To set `request_timeout=0` you have to use `Search.params()` like so: `Search.params(request_timeout=0).delete()`

or if you want to [execute the query asynchronously](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-task-api) you can call with `Search.params(wait_for_completion=False).delete()`. This will respond with a Task which you can check the status of using the [`Elasticsearch.tasks.*` APIs](https://elasticsearch-py.readthedocs.io/en/7.10.0/api.html#tasks).

Let me know if this doesn't solve your problem

Issue #1115, opened 2019-02-04
Search.delete needs to handle kwargs


ericandrewmeadows:
Search.delete does not take in keywords, such as conflicts=proceed, which is essential when trying to delete multiple objects, but currently the issue returned is:

ConflictError(409, u'{"took":1,"timed_out":false,"total":2,"deleted":0,"batches":1,"version_conflicts":2,"noops":0,"retries":{"bulk":0,"search":0},"throttled_millis":0,"requests_per_second":-1.0,"throttled_until_millis":0,"failures":[{"index":"mlsdata_test","type":"doc","id":"DlyKuWgBOtcMH22wMIjN","cause":{"type":"version_conflict_engine_exception","reason":"[doc][DlyKuWgBOtcMH22wMIjN]: version conflict, current version [2] is different than the one provided [1]","index_uuid":"YY-DGvIkSe6bGSQsobnYCA","shard":"0","index":"mlsdata_test"},"status":409},{"index":"mlsdata_test","type":"doc","id":"D1yKuWgBOtcMH22wMIjv","cause":{"type":"version_conflict_engine_exception","reason":"[doc][D1yKuWgBOtcMH22wMIjv]: version conflict, current version [2] is different than the one provided [1]","index_uuid":"YY-DGvIkSe6bGSQsobnYCA","shard":"0","index":"mlsdata_test"},"status":409}]}')
Traceback (most recent call last):
  File "<stdin>", line 186, in upsert_json_mls_listing_to_document
  File "<stdin>", line 147, in handle_previously_matched_record
  File "/mnt/yarn/usercache/livy/appcache/application_1548450731252_0176/spark-eb780ca6-de11-4c8f-a6d1-449f4b5c3168/userFiles-2fa51379-ad00-48df-a497-ede1424836b5/elasticsearch_dsl.zip/elasticsearch_dsl/search.py", line 742, in delete
    **self._params
  File "/mnt/yarn/usercache/livy/appcache/application_1548450731252_0176/spark-eb780ca6-de11-4c8f-a6d1-449f4b5c3168/userFiles-2fa51379-ad00-48df-a497-ede1424836b5/elasticsearch.zip/elasticsearch/client/utils.py", line 76, in _wrapped
    return func(*args, params=params, **kwargs)
[...]

PR #1970, opened 2025-03-27
Add param passing notes to documentation on Search.delete

lunalucadou:
Earlier today I was trying to pass `wait_for_completion=True` to `Search.delete` but found `Search.delete` does not support `**kwargs`.

I was going to just add support for `**kwargs` (it would be a fairly trivial change), but after seeing #1115 and #1395, it seems the preferred way is to call `Search.params` first. However, this is not called out in the documentation for `Search.delete`, so I instead opted to add it to the documentation.

While I was at it, I noticed `Search.scan` and `FacetedSearch.params` had broken links so I fixed those as well.

Edit: Closing in favor of MR on the other repo: elastic/elasticsearch-py#2861

### DIFF
--- a/docs/search_dsl.rst
+++ b/docs/search_dsl.rst
@@ -98,6 +98,7 @@ explicitly:
 
 Delete By Query
 ~~~~~~~~~~~~~~~
+
 You can delete the documents matching a search by calling ``delete`` on the ``Search`` object instead of
 ``execute`` like this:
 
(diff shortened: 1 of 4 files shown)


miguelgrinberg:
Thanks!

First of all, when the CLA checker doesn't detect you have signed it means that you are using a different email address on this PR than the one you used to sign.

Second, note that we are not going to continue the development of this package here. Instead, starting from the 8.18 release due in April this package is going to be deprecated and all the DSL code will be integrated with the main Python client (https://github.com/elastic/elasticsearch-py).

Would you like to submit this change over there? The DSL files are all in the `elasticsearch/dsl` subdirectory, with the same structure as in here. I can also apply the change myself if you prefer.

lunalucadou:
Ah, I have a different email in my git config for an internal gitlab instance. I'll redo the CLA and port the changes to the new repo.

PR #1979, opened 2025-07-08
Use `collections.abc.Mapping` instead of `collections.Mapping`

aalekhpatel07:
For anyone running an older version of ElasticSearch `5.x`, who wants to upgrade to `Python >=3.10` (`3.12` in my case), the current `collections.Mapping` import becomes a hard error as it has been deprecated since 3.3 and removed in 3.10.

This PR updates the imports to point to `collections.abc.Mapping` instead, and hopes to release a `5.x` minor version that works with Python3.10+.

The deprecation warning in Python3.9:
```
Python 3.9.23 (main, Jun  4 2025, 00:00:00)
[GCC 15.1.1 20250521 (Red Hat 15.1.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import collections
>>> collections.Mapping
<stdin>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
<class 'collections.abc.Mapping'>
>>>
```

The hard error in Python3.12:
```
Python 3.12.11 (main, Jul  1 2025, 15:29:55) [GCC 8.5.0 20210514 (Red Hat 8.5.0-26)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import collections.abc
>>> collections.abc.Mapping
<class 'collections.abc.Mapping'>
>>> import collections
>>> collections.Mapping
AttributeError: module 'collections' has no attribute 'Mapping'
```

### DIFF
--- a/elasticsearch_dsl/aggs.py
+++ b/elasticsearch_dsl/aggs.py
@@ -1,4 +1,4 @@
-import collections
+import collections.abc
 
 from .utils import DslBase
 from .response.aggs import BucketData, FieldBucketData, AggResponse, TopHitsData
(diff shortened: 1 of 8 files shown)


pquentin:
Thanks for opening a pull request, but as mentioned in https://github.com/elastic/elasticsearch-dsl-py/issues/1980#issuecomment-3051491394 I'm going to have to decline it, sorry.

